Issue #34, opened 2026-07-13
Unable to deploy to Cloudflare


emilpriver:
Unable to deploy to clouflare using 
```deployment: {
    output: "server",
    adapter: "cloudflare",
    site: "https://internal-docs.domain.io",
  },
```

getting errors such as
```
 prerendering static routes 
22:51:27.702
[error] Failed to get static paths from the Cloudflare prerender server (500: Internal Server Error).
22:51:27.702
Error: No such module "node:path".
22:51:27.702
  imported from "dist/server/.prerender/chunks/_.._p0PQW4lk.mjs"
22:51:27.702
    at async Object.fetch (file:///opt/buildhome/repo/node_modules/.pnpm/miniflare@4.20260708.1/node_modules/miniflare/dist/src/workers/core/entry.worker.js:4687:22)
22:51:27.703
  at Object.getStaticPaths (node_modules/.pnpm/@astrojs+cloudflare@14.1.3_astro@7.0.8_@astrojs+markdown-remark@7.2.1_@emnapi+core@1.11.1_@em_gxvc6tbm7ukgjlnqa55ut4hz7a/node_modules/@astrojs/cloudflare/dist/prerenderer.js:72:15)
22:51:27.703
  at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
22:51:27.703
  at async generatePages (node_modules/.pnpm/astro@7.0.8_@astrojs+markdown-remark@7.2.1_@emnapi+core@1.11.1_@emnapi+runtime@1.11.2_@vercel_psxiimpy4k72f2rcshppznhzt4/node_modules/astro/dist/core/build/generate.js:68:29)
22:51:27.703
  at async BasicMinimalPluginContext.handler (node_modules/.pnpm/astro@7.0.8_@astrojs+markdown-remark@7.2.1_@emnapi+core@1.11.1_@emnapi+runtime@1.11.2_@vercel_psxiimpy4k72f2rcshppznhzt4/node_modules/astro/dist/core/build/static-build.js:123:11)
22:51:27.703
  at async Object.buildApp (node_modules/.pnpm/vite@8.1.4_esbuild@0.28.1_jiti@2.7.0_yaml@2.9.0/node_modules/vite/dist/node/chunks/node.js:33427:5)
22:51:27.703
[...]

PR #35, opened 2026-07-13
Potentional fix for deploying on cloudflare

emilpriver:
This PR aims to solve #34 where I was unable to deploy to cloudflare

Update 14 july 2026: Since it looks like we're wrapping building astro inside of a CLI(which is not wrong) does it look as if we're missing wrangler.json/toml/jsonc file which is used when you build and push files to Cloudflare to for instance add nodejs compability flag or other bindings. I think it's a good idea to include this wrangler config file during build 

- [x] I've reviewed my code
- [x] I've written tests

### DIFF
--- a/.changeset/prerender-cloudflare-node.md
+++ b/.changeset/prerender-cloudflare-node.md
@@ -1,0 +1,5 @@
+---
+"blume": patch
+---
+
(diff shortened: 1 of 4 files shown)


emilpriver:
@haydenbleasel I was looking into docs but don't know how to write it since the goal is to wrap as much as possible into `blume` cli

haydenbleasel:
Thank you so much for this, @emilpriver! 🙏 This was a genuinely thorough fix — tracking the breakage down to Astro 6 switching the `@astrojs/cloudflare` prerender runtime from Node to workerd, and wiring both `prerenderEnvironment: "node"` and the `wrangler` `configPath` through the generated config, is exactly the right shape. The changeset write-up and the tests made it really easy to review.

I pushed one tiny follow-up commit to simplify an unreachable branch in the `configPath` normalization (and drop the now-unused `basename` import) — nothing that changes behavior.

Really appreciate you digging into #34 and seeing it through. Merging now! 🚀

PR #37, opened 2026-07-13
 use rounded-blume token in PageFeedback and Pagination

### LINKED ISSUE #36: Pagination and PageFeedback ignore theme radius
The `Pagination` and `PageFeedback` components hardcode `rounded-full` on their buttons, bypassing the theme's `radius` option that we set in blume.config.ts.

**Reproduce:**

1. Set `theme.radius: "md"` (or any value other than the `full` shape) in `blume.config.ts`
2. Navigate to any page with previous/next pagination or the "Was this page helpful?" feedback section
3. The buttons render as fully rounded pills instead of using the configured radius

**Root cause:**
Both Pagination.astro or PageFeedback.astro seem to be using a hard-coded rounded-full for their styles instead of `rounded-blume`, which is the utility that maps to `--blume-radius` (the CSS variable from the theme config).

**Expected behavior:**

These components should use `rounded-blume` so they respect the theme's `radius` setting, consistent with the rest of Blume's UI (cards, badges, callouts, etc.).

**Environment:**
- Blume: 1.0.3
- Node: 22.x

**Workaround:**

Override both via layout slots in `components.ts`:

```ts
import { defineComponents } from "blume";
import Pagination from "./components/Pagination.astro";
import PageFeedback from "./components/PageFeedback.astro";

export default defineComponents({
  layout: {
    Pagination,
    Feedback: PageFeedback,
  },
});
```

Then copy the built-in components and replace `rounded-full` with `rounded-blume`.

0xherve:
## Description

Replace hardcoded rounded-full with the rounded-blume design token in `PageFeedback` and `Pagination` components, so that they respect the configurable --blume-radius theme variable.

## Related Issues

Closes #36 

## Checklist

- [x] I've reviewed my code
- [ ] I've written tests

The fix ensures custom radius values are respected

## Additional Notes

There are additional components still using hardcoded rounded-full that should also use rounded-blume (e.g. Search, LanguageSwitcher, Banner). Happy to commit those fixes to address those two.

### DIFF
--- a/.changeset/rounded-blume-feedback-pagination.md
+++ b/.changeset/rounded-blume-feedback-pagination.md
@@ -1,0 +1,7 @@
+---
+"blume": patch
+---
+
+Use the `rounded-blume` radius token instead of a hardcoded `rounded-full` on the `PageFeedback` and `Pagination` buttons so they respect the configured `theme.radius`.
+
+Fixes #36.
--- a/packages/blume/src/components/layout/PageFeedback.astro
+++ b/packages/blume/src/components/layout/PageFeedback.astro
@@ -15,7 +15,7 @@ const { strings } = Astro.props;
 const f = { ...EN_UI.feedback, ...strings };
 
 const buttonClass =
-  "inline-flex items-center gap-2 rounded-full border border-border px-4 py-2 text-foreground text-sm transition-colors hover:border-foreground";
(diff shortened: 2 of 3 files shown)


haydenbleasel:
Thanks so much for this, @0xherve! 🙌 Clean, well-scoped fix that maps these controls onto the `rounded-blume` token exactly the way the rest of the UI works. I went ahead and pushed a changeset onto the branch so it flows into the next release. Really appreciate the clear write-up and the heads-up about the other `rounded-full` controls — merging now. 🌸

PR #45, opened 2026-07-14
Wrap markdown tables in a horizontal-scroll container

SutuSebastian:
## Summary
- Wide markdown tables overflow the x-axis because the built-in table render has no scroll wrapper.
- Adds a `blume:table-wrap` hast plugin that wraps every `<table>` in `<div class="blume-table-scroll">…</div>`, registered in the shared hast pipeline so it covers both `.md` and `.mdx`.
- Adds `.blume-table-scroll { overflow-x: auto; -webkit-overflow-scrolling: touch }` to the theme.
- The visitor returns a replacement node; Satteri does not re-descend into it, so the wrapped table is not re-visited (no double-wrap).
- Adds a changeset (`blume` patch).

## Test plan
- [x] `bun test packages/blume/test/table-wrap.test.ts` — new unit test asserts the visitor wraps a `table` in the scroll div and reuses the original node.
- [x] `bun run typecheck` (blume package) — clean.
- [x] `oxfmt --check` / `oxlint --quiet` on changed files — clean.
- [ ] Build a docs page with a 7+ column table; confirm it scrolls in place instead of overflowing.

### DIFF
--- a/.changeset/wide-tables-scroll.md
+++ b/.changeset/wide-tables-scroll.md
@@ -1,0 +1,5 @@
+---
+"blume": patch
+---
+
+Wrap markdown tables in a horizontal-scroll container so wide tables no longer overflow the prose column.
--- a/.changeset/wide-tables-scroll.md
+++ b/.changeset/wide-tables-scroll.md
@@ -2,4 +2,4 @@
 "blume": patch
 ---
 
-Wrap markdown tables in a horizontal-scroll container so wide tables no longer overflow the prose column.
+Wrap markdown tables in a horizontal-scroll container so wide tables no longer overflow the prose column. The wrapper is framed with a rounded border and cell padding, and header labels stay on one line, so a clipped wide table reads as scrollable rather than cut off.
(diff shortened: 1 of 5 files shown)


haydenbleasel:
Thanks for this! I pushed a couple of follow-up commits that build on the scroll wrapper.

Tested it against a 10-column table in the docs and a bare `overflow-x: auto` wrapper works, but the hard clip at the column edge reads as "cut off" rather than "scrollable." So the wrapper is now framed:

- **Border + radius** on `.blume-table-scroll` (`--blume-border` / `--blume-radius`) so the clip looks like a scroll viewport.
- **Cell padding** — Tailwind Typography zeroes the first/last cell's inline padding (so borderless tables align to the prose margin), which left edge text touching the new border; `:is(th, td)` padding restores it and outweighs Typography's `:where()`-scoped rules.
- **`white-space: nowrap` on `th`** so two-word labels (e.g. "Start Date") don't wrap into a ragged stack; body cells still wrap.
- **Keyboard a11y** — the wrapper now carries `tabindex="0"` so a horizontally scrolling table is reachable and scrollable by keyboard, not just pointer (WCAG 2.1.1 / axe `scrollable-region-focusable`). Added a unit test for it.
- Updated the changeset wording to cover the framing.

(Tried a muted header background too but it clashed with the rest of the component set, so left it off.)

SutuSebastian:
Even better, thanks!

PR #51, opened 2026-07-14
Keep root tabs active on child routes

### LINKED ISSUE #49: Root navigation tab is inactive on child routes
When a tab uses `/` as the default section, it is only active on the home page. Child routes such as `/guides/services` show no active tab.

Treat `/` as the fallback match and select the longest matching tab, so more specific sections still win.

radjavi:
## Summary
- use one longest-path tab matcher across desktop and mobile navigation
- treat the root tab as the fallback when no more specific section matches
- add focused tests and a patch changeset

Closes #49.

## Test plan
- [x] `bun run check`
- [x] `bun run typecheck`
- [x] `bun run build`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/root-tab-child-routes.md
+++ b/.changeset/root-tab-child-routes.md
@@ -1,0 +1,5 @@
+---
+"blume": patch
+---
+
+Keep root navigation tabs active on descendant routes unless a more specific tab matches.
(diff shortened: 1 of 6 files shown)


haydenbleasel:
Thanks @radjavi! Nice fix — unifying the tab matcher into `activeTabForRoute` cleans up the duplicated `aria-current` logic and also fixes the nested-tab case where two tabs could both be marked current. Reviewed and merging. 🙏

PR #53, opened 2026-07-14
Prevent themed logo layout shifts

### LINKED ISSUE #52: Per-mode logo images shift the header while loading
Per-mode logos render as external `` elements with a fixed height but no intrinsic width or height. Until the image metadata loads, the browser cannot reserve the final width. Once it loads, the logo, tabs, and search move.

Single SVG logos are inlined and do not have this problem. For local SVG variants, read each file's dimensions and pass them to the rendered images.

radjavi:
## Summary
- read dimensions from local light and dark SVG logos at build time
- emit intrinsic dimensions on logo images so the header stays stable before load
- cover explicit dimensions and view boxes in tests

Closes #52.

## Test plan
- [x] `bun run check`
- [x] `bun run typecheck`
- [x] `bun run build`

### DIFF
--- a/.changeset/logo-layout-shift.md
+++ b/.changeset/logo-layout-shift.md
@@ -1,0 +1,5 @@
+---
+"blume": patch
+---
+
+Reserve space for local per-mode SVG logos before they load.
--- a/packages/blume/src/astro/generate.ts
+++ b/packages/blume/src/astro/generate.ts
@@ -626,7 +626,56 @@ const writeStagedContent = async (
   }
 };
 
-/** The logo shape the runtime consumes: an inline SVG or image URL(s). */
+interface LogoDimensions {
+  height: number;
+  width: number;
+}
+
+const SVG_ROOT = /<svg\b(?<attributes>[^>]*)>/u;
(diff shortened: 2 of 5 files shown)


haydenbleasel:
Nice work — this is a clean, well-scoped fix. Reading the intrinsic dimensions at build time and stamping `width`/`height` on the per-mode `` tags is exactly the right way to reserve space and kill the horizontal shift, and it's good to see both the explicit-size and `viewBox` paths covered by a test. The single-inline-SVG and "file absent → fall back to ``" behaviors are both preserved.

One small edge case for a follow-up (not blocking): `SVG_WIDTH` / `SVG_HEIGHT` use `\bwidth`/`\bheight`, and `\b` matches the boundary in `stroke-width`, so a root `<svg>` like `<svg stroke-width="2" width="200" height="50">` reads the width as `2`. It's order-dependent within the attribute list, so real-world impact is low, but a negative lookbehind would make it robust:

```js
const SVG_WIDTH = /(?<![-\w])width\s*=\s*["'](?<value>[^"']+)["']/u;
const SVG_HEIGHT = /(?<![-\w])height\s*=\s*["'](?<value>[^"']+)["']/u;
```

Merging now — thanks for this! 🙏

PR #55, opened 2026-07-14
Add Open Graph card branding

### LINKED ISSUE #54: Generated Open Graph cards cannot be branded
Generated Open Graph cards only inherit the theme accent. Their background, foreground, muted text, and border colors are hardcoded. Per-mode logos also fall back to the initial tile because the renderer only receives a single inlined SVG.

Add optional `logo` and `palette` settings under `seo.og`, while keeping the current card unchanged when they are omitted.

radjavi:
## Summary
- add optional local SVG and hex palette settings under `seo.og`
- pass resolved branding into generated cards while preserving current defaults
- document the config and cover schema, generation, templates, and rendering

Closes #54.

## Test plan
- [x] `bun run check`
- [x] `bun run typecheck`
- [x] `bun run build`
- [x] render a card with a custom logo and palette

### DIFF
--- a/.changeset/og-card-branding.md
+++ b/.changeset/og-card-branding.md
@@ -1,0 +1,5 @@
+---
+"blume": minor
+---
+
+Add custom logo and palette settings for generated Open Graph cards.
--- a/apps/docs/content/docs/configuration/seo.mdx
+++ b/apps/docs/content/docs/configuration/seo.mdx
@@ -106,7 +106,26 @@ seo: {
 }
 ```
 
-Each card is derived from your content and theme — the **page title** as the headline, your **site title** as the eyebrow, and your theme **accent** for the mark. Images are served at `/og/<slug>.png`, mirroring each route, and are prerendered as static files even in server mode:
+### Brand the generated card
+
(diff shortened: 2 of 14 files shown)


haydenbleasel:
Thanks @radjavi — this is a clean, well-tested addition. The threading from config → schema → runtime data → renderer is tidy, the defaults are fully preserved, and I like the defensive hex validation at both the schema and render layers.

We discussed whether this overlaps with the existing `theme` config, since the card already reads `theme.accent`. The conclusion: `blume.config.ts` doesn't yet model a full light/dark theme (no `foreground`/`muted`/`border` slots), and the OG card is a deliberately mode-independent, hex-only raster surface that Takumi renders separately — so a small dedicated palette is the right call for now. If we ever grow a fuller theme concept, we can revisit deriving these automatically.

I resolved a merge conflict against `main` (the `referenceTabs` → `referenceRoutes` rename) and downgraded the changeset from `minor` to `patch` to match how we're versioning during the pre-1.0 window. Typecheck and the full suite pass. Merging now — thanks again! 🌸

PR #57, opened 2026-07-14
Author API reference tabs via navigation.tabs (drop auto-generated tabs) + fix Example API link

jaydenfyi:
## Description

API references (OpenAPI/AsyncAPI) no longer add a header tab automatically. Instead, you author the tab yourself by pointing a `navigation.tabs` entry at the reference route:

```ts blume.config.ts
navigation: {
  tabs: [{ label: "API", path: "/reference" }],
}
```

This gives you control over the label and placement, and — for the native renderer — a tab at the reference route is what scopes its operations sidebar. Making the tab author-defined keeps that behavior consistent: a tab means a scoped section; no tab means the pages stay reachable at their routes. Reference routes are still whitelisted as valid navigation targets, so an authored tab pointing at one doesn't read as a broken link.

Also fixes the broken **Example API** link on the API reference docs page — that `/api` route was intentionally removed earlier, so the docs now describe the config example instead of linking to a live page.

## Changes

- References no longer auto-generate a header tab; `navigation.generatedTabs` is not needed and was removed.
- `referenceTabs` → `referenceRoutes` (used for nav-target validation).
- Docs updated (`navigation`, `api-reference`) to describe the author-defined tab pattern.
- Removed the dead `[Example API](/api)` link.

## Checklist

- [x] I've reviewed my code
- [x] I've written tests

### DIFF
--- a/.changeset/fresh-rivers-rest.md
+++ b/.changeset/fresh-rivers-rest.md
@@ -1,0 +1,5 @@
+---
+"blume": patch
+---
+
+Allow framework-generated header tabs to be disabled with `navigation.generatedTabs`.
--- a/.changeset/fresh-rivers-rest.md
+++ b/.changeset/fresh-rivers-rest.md
@@ -2,4 +2,4 @@
 "blume": patch
 ---
 
-Allow framework-generated header tabs to be disabled with `navigation.generatedTabs`.
+API references (OpenAPI, AsyncAPI) no longer add a header tab automatically. Point a `navigation.tabs` entry at the reference route to surface it — this also lets you control the tab label and scopes the operations sidebar for the native renderer.
--- a/apps/docs/blume.config.ts
+++ b/apps/docs/blume.config.ts
@@ -1,6 +1,9 @@
 import { defineConfig } from "blume";
 
 export default defineConfig({
+  ai: {
+    llmsTxt: { openapi: false },
+  },
   analytics: {
     vercel: true,
   },
@@ -35,9 +38,20 @@ export default defineConfig({
   lastModified: true,
   logo: "/logo.svg",
   navigation: {
+    generatedTabs: false,
     tabs: [
       { label: "Docs", path: "/docs" },
       { label: "Changelog", path: "/changelog" },
+    ],
+  },
+  openapi: {
+    enabled: true,
+    route: "/api",
+    sources: [
+      {
+        label: "Example API",
+        spec: "https://petstore3.swagger.io/api/v3/openapi.json",
+      },
     ],
   },
   seo: {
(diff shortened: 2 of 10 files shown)


haydenbleasel:
Thanks for catching the broken link! I pushed a commit (c4953bf8) that adjusts the docs-site side of this.

Context: the "Example API" petstore reference on the docs site was **intentionally removed** in e523f245 — a pet store spec sitting on the Blume docs site is confusing, since it documents the Petstore, not Blume. So re-enabling that `openapi` block here is an accidental revert of that decision, and I've reverted it back out (along with the `ai.llmsTxt` and `navigation.generatedTabs: false` that only existed to support it).

The actual leftover behind the broken link was a stale sentence in `advanced/api-reference.mdx` that still linked to `[Example API](/api)` — which 404s now that `/api` no longer exists. I reworded it to just describe the config example instead of linking to a live page, so the link is fixed without putting the pet store back on the site.

One more thing worth knowing for the feature itself: `generatedTabs: false` doesn't fully hide a **Blume-rendered** reference (the default renderer). Since Blume renders each operation as a real page in the sidebar, dropping the header tab just relocates those operations into the untabbed/root sidebar rather than hiding them — I confirmed this shows the whole petstore operations group on the homepage sidebar. Scalar-rendered refs hide cleanly. Might be worth deciding whether `generatedTabs: false` should also prune the reference's sidebar section for the Blume renderer.

I've left the `generatedTabs` feature itself (schema, generate.ts, changeset, navigation docs, tests) untouched — that's a reasonable general capability, and I'll follow up on the rest.

haydenbleasel:
Followed up with a different approach for the tab side of this (e31a5781).

Rather than auto-generate the reference tab and add a `generatedTabs` flag to suppress it, I dropped the automatic tab entirely — API references now mount at their route but don't add a header tab on their own. To surface one, you point a `navigation.tabs` entry at the reference route:

```ts
navigation: { tabs: [{ label: "API", path: "/reference" }] }
```

Two reasons this is cleaner than the toggle:

- **You control the label and placement** — it's just a normal tab, no special-casing.
- **It fixes the sidebar behavior** instead of working around it. The tab is what scopes a native-rendered reference's operations sidebar, so making the tab author-defined means "has a tab → scoped section, no tab → loose pages" is one consistent rule. The `generatedTabs: false` path couldn't express that — it hid the tab but left the operations in the sidebar.

Changes: `referenceTabs` → `referenceRoutes` (routes still whitelisted as valid nav targets so an authored tab doesn't warn), `generatedTabs` option removed, docs + tests updated. Kept it a `patch`. Full suite + coverage green.

Thanks again for kicking this off — the broken-link fix and the nudge to rethink reference tabs were both spot on.

jaydenfyi:
@haydenbleasel Awesome. Yeah I quickly realised after opening the PR it wasn't just as straight forward as fixing a broken link haha.

Definitely think your approach is the more intuitive behaviour, was just trying to play it safe with backwards compatibility of the current config 😅

PR #58, opened 2026-07-14
Hide language icon on header-less standalone code blocks

### LINKED ISSUE #56: Language icon overlaps code when a CodeBlock has no title
A `<CodeBlock>` without a `title` renders the language icon (e.g. the TS badge) on top of the first line of code.

## Why

Two transformers are out of sync:

- `languageIconTransformer` runs whenever `icons !== false` (default on) — it injects the `<svg class="blume-lang-icon">` and sets `data-icon` on the `<pre>`.
- `languageAttrTransformer` (which sets `data-language`) only runs when a `title` is passed (`src/markdown/index.ts`, ~the "Applied only on the titled path" comment).

The theme's header space (`padding-top: 3.75rem` on `pre[data-language]`) and header bar (`::before`) key off `data-language`. So without a title there's no header padding, and the absolutely-positioned icon (`top: 0.875rem`) lands on line 1.

## Repro

```mdx
<CodeBlock lang="ts" code={`export const x = 1;`} />
```

Icon overlaps `export`. Add `title="foo.ts"` and it renders fine (header bar + padding show up).

## Scope

`<CodeBlock>` and direct `highlightCode` callers without a title. MDX fences are fine — Astro's markdown Shiki sets `data-language` on those.

## Suggested fix

Either gate the icon on having a title (mirror `languageAttrTransformer`), or always set `data-language` so the header padding is reserved whenever the icon shows. The latter feels safer — the icon and the header it lives in should never be separable.

## Source

- `src/markdown/index.ts` — icon transformer pushed unconditionally (~L97-99); `languageAttrTransformer` title-gated (~L128-134)
- `src/markdown/language-icon.ts` — sets `data-icon` + unshifts the svg
- `src/theme/entry.ts` — `pre[data-language]` padding/header (~L362-407)

SutuSebastian:
## Summary

- Gates the language-icon display rule on `data-language` (the header bar) instead of just `data-icon`, so the icon no longer overlaps the first code line on a header-less standalone `<CodeBlock>`.
- The icon transformer runs whenever icons are on, but a standalone `<CodeBlock>` without a `title` never gets `data-language` (the header that reserves the icon's space via `padding-top`), so the absolutely-positioned icon sat on top of the first line. Fenced code gets `data-language` from Shiki automatically, and titled blocks get it from `languageAttrTransformer` — both keep the header, so the icon still shows there.
- One-line CSS change in `packages/blume/src/theme/entry.ts` (two selectors gain a `data-language` qualifier), plus a regression test.

Fixes #56.

## Root cause

- `languageIconTransformer` (`packages/blume/src/markdown/language-icon.ts`) injects the `<svg class="blume-lang-icon">` and sets `data-icon` on `<pre>` whenever `icons !== false` — independent of whether a title/header exists.
- The header bar that reserves room for the icon is keyed off `pre[data-language]` (`packages/blume/src/theme/entry.ts`): `pre[data-language] { padding-top: 3.75rem }` and the `::before` label.
- On the standalone `highlightCode` path, `data-language` is only added by `languageAttrTransformer`, which runs **only when a `title` is passed** (`packages/blume/src/markdown/index.ts`). Fenced code is unaffected because Shiki sets `data-language` itself.
- Net effect: header-less standalone block → `data-icon` set, `data-language` absent → no header padding → the absolutely-positioned icon (`top: 0.875rem`) overlaps the first code line.

## Fix

Scope the icon display (and the label `padding-left` shift) to `pre[data-language][data-icon]`, so the icon renders only when a header bar exists to hold it. This matches the existing comment's stated intent ("Shown only for top-level prose blocks, which have a header") and leaves fenced code, titled blocks, tabs, and API panels untouched.

```diff
-.prose > :where(pre[data-icon]) > .blume-lang-icon {
+.prose > :where(pre[data-language][data-icon]) > .blume-lang-icon {
   ...
 }
-.prose > :where(pre[data-icon])::before {
+.prose > :where(pre[data-language][data-icon])::before {
   padding-left: 2.5rem;
 }
```

## Test plan

- [x] `bun run test` — 1399 pass, 0 fail (includes a new regression test asserting the header-less path omits `data-language` while still emitting `data-icon`/`blume-lang-icon`).
- [x] `bun run check` — formatting clean.
- [x] `bun run build` (via pre-commit/turbo) — succeeds.
- [x] Reviewer: confirm a header-less `<CodeBlock lang="ts" code="..." />` no longer shows an overlapping icon, and that fenced + titled blocks still render the icon.

### DIFF
--- a/.changeset/codeblock-icon-overlap.md
+++ b/.changeset/codeblock-icon-overlap.md
@@ -1,0 +1,7 @@
+---
+"blume": patch
+---
+
+Gate the language icon on `data-language` so it no longer overlaps the first code line on header-less standalone `<CodeBlock>`s. The icon transformer runs whenever icons are on, but a standalone block without a `title` never gets `data-language` (the header bar that reserves the icon's space), so the absolutely-positioned icon sat on top of the first line. Fenced code and titled blocks are unaffected.
+
+Fixes #56.
--- a/apps/docs/content/docs/content/components.mdx
+++ b/apps/docs/content/docs/content/components.mdx
@@ -619,7 +619,6 @@ An Astro example renders live with no client JavaScript:
 
 <CodeBlock
   lang="ts"
-  icons={false}
   code={`export const greet = (name: string): string =>
   \`Hello, \${name}!\`;`}
 />
@@ -629,7 +628,7 @@ An Astro example renders live with no client JavaScript:
 import CodeBlock from "blume/components/content/CodeBlock.astro";
 ---
 
-<CodeBlock lang="ts" code={source} icons={false} />
+<CodeBlock lang="ts" code={source} />
 ```
 
 To highlight to an HTML string yourself (e.g. inside your own component), import the underlying helper from `blume/markdown`:
--- a/packages/blume/src/markdown/language-icon.ts
+++ b/packages/blume/src/markdown/language-icon.ts
@@ -7,7 +7,8 @@
  * hex colors are skipped because dark-on-dark logos (Next.js, Rust…) vanish.
  *
  * The theme styles `.blume-lang-icon` and shifts the language label
- * (`pre[data-icon]::before`) to make room.
+ * (`pre[data-language][data-icon]::before`) to make room — gated on
+ * `data-language` so the icon only shows when a header bar exists to hold it.
  */
 
 import {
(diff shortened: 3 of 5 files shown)


haydenbleasel:
Thanks for the clean, well-documented fix @SutuSebastian! Root-cause writeup was spot on. I pushed two small follow-ups: dropped the now-unnecessary `icons={false}` workaround from the CodeBlock docs example, and synced the transformer comment with the new gate. Merging. 🙏

SutuSebastian:
Glad I could help!

PR #59, opened 2026-07-14
Match native controls to active theme

### LINKED ISSUE #40: Sidebar scrollbar appears light in dark mode
**Describe the bug** A clear and concise description of what the bug is.

On macOS, when a mouse is connected and dark mode is enabled, the sidebar scrollbar is displayed using the light theme instead of the dark theme.

**Expected behavior** browser can handle.
```css
html.dark {
  color-scheme: dark;
}

html{
  color-scheme: light;
}
```

Image

FranciscoMoretti:
## Summary

- match native browser controls, including scrollbars, to Blume's active color theme
- emit the shared `:root` color-scheme defaults in both the docs shell and isolated example previews
- use Blume's existing `data-theme` contract instead of introducing a custom scroll area
- add focused regression coverage and a patch changeset

Fixes #40.

## Root cause

Blume updates `data-theme` on the root element, but the generated theme stylesheets did not set `color-scheme`. Native controls therefore continued using the browser's default light appearance in dark mode.

### Before

Light scrollbar in dark mode

### After — dark theme

Dark scrollbar matching dark theme

### After — light theme

Light scrollbar matching light theme

## Validation

- `bun run check`
- `bun run typecheck`
- `bun run test:coverage` — 1,400 passed, 0 failed
- `bun run build --filter=!video --filter blume`
- verified the overflowing sidebar renders with `color-scheme: dark` in dark mode
- toggled to light mode and verified `color-scheme: light`
- no browser console warnings or errors

### DIFF
--- a/.changeset/quiet-scrollbars-smile.md
+++ b/.changeset/quiet-scrollbars-smile.md
@@ -1,0 +1,5 @@
+---
+"blume": patch
+---
+
+Match native browser controls, including scrollbars, to the active color theme.
--- a/packages/blume/src/theme/entry.ts
+++ b/packages/blume/src/theme/entry.ts
@@ -159,9 +159,13 @@ ${THEME_MAPPING}
     cursor: pointer;
   }
   html {
+    color-scheme: light;
     scroll-behavior: smooth;
     scroll-padding-top: 4.5rem;
     text-rendering: optimizeLegibility;
+  }
+  html[data-theme="dark"] {
+    color-scheme: dark;
   }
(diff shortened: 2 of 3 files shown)


haydenbleasel:
Thanks for the fix, @FranciscoMoretti! Clean, well-scoped, and exactly the right approach — leaning on the existing `data-theme` contract with `color-scheme` rather than a custom scroll area. Both sheets covered, tests and changeset included. Merging. 🙌

Issue #62, opened 2026-07-15
OG image generation renders CJK text as tofu — Takumi has no font-loading config surface


h-kono-it:
**Is your feature request related to a problem? Please describe.**

Blume's per-page Open Graph image (rendered via Takumi in `og/card.ts`) always renders CJK text (Japanese, Chinese, Korean, ...) as tofu (missing-glyph boxes). Takumi's built-in last-resort font (Geist) only covers Latin glyphs, and there's currently no config option anywhere in `blume.config.ts` to load an additional font into the `Renderer` instance used for OG cards. This affects any docs site whose title/description isn't in a Latin script.

<img width="1200" height="630" alt="Image" src="https://github.com/user-attachments/assets/883cb5b7-6b49-4f43-ba6e-8b19ec9d7739" />

**Describe the solution you'd like**

Expose a config option to load one or more custom fonts into the OG renderer, e.g.:

```ts
seo: {
  og: {
    fonts: [
      "assets/og-fonts/NotoSansJP-Regular.otf",
      { path: "assets/og-fonts/NotoSansJP-Bold.otf", weight: 700 },
    ],
  },
}
```

Takumi's `Renderer` already supports this at the engine level (`loadFontSync`/`loadFonts`/the constructor's `fonts` option) and does per-glyph fallback across every loaded font — "a missing glyph walks the chain until a family covers it" — so this wouldn't touch how Latin text renders today; it would just extend the fallback chain `card.ts`'s `getRenderer()` builds.

I validated this works: patching `getRenderer()` locally to `loadFontSync` a bundled Noto Sans JP (Regular + Bold) alongside the default renderer fixes it completely, with zero change to `text()` calls in `card.ts`.

```sh
// Proof-of-concept patch to og/card.ts's getRenderer():
const loadRegionalOgFonts = (target: Renderer): void => {
  target.loadFontSync({
    data: readFileSync("assets/og-fonts/NotoSansJP-Regular.otf"),
    name: "Noto Sans JP",
    weight: 400,
  });
  target.loadFontSync({
    data: readFileSync("assets/og-fonts/NotoSansJP-Bold.otf"),
    name: "Noto Sans JP",
    weight: 700,
  });
};

let renderer: Renderer | null = null;
const getRenderer = (): Renderer => {
  if (!renderer) {
    renderer = new Renderer();
    loadRegionalOgFonts(renderer);
  }
  return renderer;
};
```

[...]

PR #63, opened 2026-07-15
Upgrade to Takumi v2

kane50613:
## Description

Upgrading to use `takumi-js` package v2, no more `Renderer` boilerplate needed.

Added an `images` option so remote image fetching can be customized. Emoji in titles now render as Twemoji glyphs, and the OG card palette accepts any CSS color (matching `theme.accent`) rather than hex only.

A `fonts` option was part of this PR originally, but it turned out to be a no-op — `takumi-js` never reads `options.fonts`, so [google fonts](https://takumi.kane.tw/docs/typography-and-fonts#from-google-fonts) don't work through it. It's been removed here and earmarked for a follow-up; see [this comment](https://github.com/haydenbleasel/blume/pull/63#issuecomment-4999284884) for the details.

## Checklist

- [x] I've reviewed my code
- [x] I've written tests

image
image

## Additional Notes

@kane50613 — I pushed review fixes to the branch and edited the Description above to drop the `fonts` claim, so the PR matches what it now ships. Everything else is yours, untouched.


haydenbleasel:
Amazing, switch this over into ready mode whenever you want me to review/merge. Appreciate it 🙏

kane50613:
> Amazing, switch this over into ready mode whenever you want me to review/merge. Appreciate it 🙏

I'm working on fixing the svg source issue to allow without `xmlns="http://www.w3.org/2000/svg"`, will switch to ready after the fix and release soon

packages/blume/src/og/card.ts
```diff
@@ -66,6 +51,10 @@ export interface OgCardOptions {
   repo?: string;
   /** Footer-right site host, e.g. `docs.acme.com`. */
   site?: string;
+  /** Fonts to be used in the card. Defaults to Geist with latin subset. */
+  fonts?: RenderOptions["fonts"];
+  /** Custom image loaders to be used in the card. */
+  images?: RenderOptions["images"];
 }
```
kane50613, line 60: I'm not sure how this should be documented or turned into even simpler shape, the goal is to allow custom fonts to solve #62

kane50613:
@haydenbleasel its ready 😁 and a comment

haydenbleasel:
Thanks for this — the v2 API is a nice cleanup, and losing the `Renderer` boilerplate is a clear win. Twemoji rendering in particular looks great; the cards are noticeably better than what v1 gave us. I've pushed a few follow-up fixes directly to the branch (hope that's alright) — here's what changed and why.

### `fonts` didn't do anything, so I stripped it

This is the one I'd most like your read on, since you know the library far better than I do.

As far as I can tell `takumi-js` never reads `options.fonts` — `render()` spreads options into `renderer.render()`, and nothing in the dist calls `registerFont`. The napi `RenderOptions` has no `fonts` field either, only `fontFamilies`. I tested it by loading Playfair Display through `googleFonts()` and passing it as `fonts`: the output was byte-identical to the default Geist render, with and without a matching `fontFamily` on the text nodes.

There may be a small type bug behind this. `RenderOptions["fonts"]` typechecks and resolves to `any[] | Promise<any[]> | undefined`, which looks like it's leaking in from the wasm side of the `napi.RenderOptions | wasm.RenderOptions` union. That's why the option looked correct in review — the compiler happily accepted a field the napi backend ignores.

I removed it rather than ship an option that silently does nothing. I did try the `Renderer` + `registerFont()` + `options.renderer` path, but couldn't get `registerFont` to accept a `googleFonts()` subset — `data` is a lazy loader function, and my attempts came back with zero registered families. I stopped rather than guess at your API.

**This is where I'd be very keen on a follow-up PR.** Google Fonts on OG cards is genuinely worth having, and Blume already carries a curated slug→family registry (`packages/blume/src/theme/fonts.ts`, driving `theme.fonts`) that a `seo.og` font option could reuse — so a card could match the site's typeface. If you're up for it, that'd be a great next one. No rush, and equally happy to take just a pointer on the right incantation if you'd rather I do the wiring.

### Emoji glyphs are now fetched once per build

v2 fetches each Twemoji glyph from the CDN per render, and we prerender one card per page — so an emoji in the site title refetched the same glyph for every page (200-page site → 200 identical requests), and any one failing fails the build. `card.ts` now passes a shared URL-keyed `images.fetchCache`, so a build fetches each glyph once. An explicit caller `fetchCache` still wins.

### `seo.og.palette` accepts any CSS color

The docs note was right in spirit but the schema disagreed: `ogColorSchema` still enforced hex-only, so `seo.og.palette.accent: "oklch(…)"` was rejected at config load even though `theme.accent` (the card's default accent) has always accepted it. Loosened the schema so both surfaces take the same grammar, and updated the note plus the test that pinned the old behavior.

Kept your fail-fast decision on malformed colors — a build error naming the value beats silently shipping a default-colored card. Also `await`ed the two `.rejects` assertions and pinned them to the offending value.

### Minor

Merged the two changesets into one and spelled out the effects, including `renderOgImage` now returning `Uint8Array` rather than `Buffer` (worth flagging in the notes for anyone importing `blume/og`).

One thing worth knowing for future OG work: our GitHub Actions build renders zero cards, because `og.enabled` defaults to `Boolean(site)` and `site` is auto-detected from `VERCEL_URL` and friends, which don't exist in Actions. So green CI says nothing about OG — the Vercel preview is the only signal that exercises it. I checked the preview and the cards render correctly on Linux.

PR #66, opened 2026-07-15
Skip headings inside <Prompt> when extracting the table of contents

### LINKED ISSUE #65: TOC leaks headings from inside <Prompt> component's hidden content
**Describe the bug**
`extractHeadings` (packages/blume/src/core/sources/normalize.ts) scans MDX as raw text with no JSX awareness. `<Prompt>` renders its slot children into a permanently `hidden` node (used only to build the copy-to-clipboard / Cursor-deeplink text) — but heading extraction doesn't know that. Any `##` line inside a `<Prompt>` block gets extracted as a real page heading and shows up in the "On this page" TOC sidebar, linking to content that never renders visibly.

**Blume version** 1.0.4

**To Reproduce**
1. Add a doc page with:
   ```
   <Prompt description="...">
   ## Setup
   ...
   </Prompt>
   ```
2. Run `blume dev` or `blume build`
3. Open the page
4. "On this page" lists "Setup" even though it never renders on screen

**Expected behavior**
Headings inside `<Prompt>`'s children should not appear in the TOC.

**Environment (please complete the following information):**
- Command: blume dev, blume build
- Runtime: Bun 1.3.14
- OS: macOS 15
- Browser (if a rendering issue): n/a (server-side extraction bug)

vasanth-rb:
## Description
`extractHeadings` treats `<Prompt>`'s children as regular page content, even though `Prompt.astro` renders them into a permanently hidden node. Any `##` inside a `<Prompt>` block leaked into the page's TOC.

Fix tracks `<Prompt>`/`</Prompt>` nesting depth the same way fenced code blocks are already tracked, and skips heading extraction while inside one. Confirmed no other content component (Tabs, Card, Panel, Expandable, Accordion, AccordionItem, Tile, Frame, TreeFolder) shares this always-hidden-children pattern, so the fix is scoped to Prompt only.

## Related Issues
Closes #65

## Checklist
- [x] I've reviewed my code
- [x] I've written tests
- [ ] I've updated the docs, if necessary — n/a, internal scanner behavior only

## Additional Notes
Verified via `bun run check`, `bun test`, `bun run typecheck`, and the pre-commit `test:coverage` gate (1420 pass, 98% threshold met).


haydenbleasel:
Thanks for the fix — the depth-tracking approach beside the existing fence tracking is exactly the right shape, and the hidden-node premise checks out. While reviewing I found three inputs where the unanchored `<Prompt` matching silently swallowed every heading for the rest of the document, so I've pushed a follow-up commit (46d04ba9) to this branch that hardens the matching:

**Issues found (each verified by running the branch code):**

1. **A multi-line self-closing tag never released the depth.** `<Prompt` + attribute lines + `/>` matched `PROMPT_OPEN` on the first line, but `PROMPT_SELF_CLOSE` only fires when the whole tag sits on one line — so the depth stuck at 1 and every later heading vanished from the TOC.
2. **A prose mention like `` `<Prompt>` `` opened a hidden region.** The unanchored regex matched inside inline code, again eating all subsequent headings on the page — one sentence of docs about the component itself would have triggered it.
3. **A heading mentioning the tag (`## The <Prompt> component`) was dropped**, along with everything after it, because the open check ran before the ATX match.

**What the follow-up changes:**

- Tag detection is now anchored to the (trimmed) start of a line — block-level JSX in MDX starts its own line, so mid-prose and mid-heading mentions can never match. A stray `</Prompt>` outside a prompt is now just paragraph text, which also removed the `Math.max` clamp.
- A new `finishPromptTag` helper + `promptTag` state carry an unterminated `<Prompt` opening tag across lines; the depth only increments once the terminating `>` proves the tag isn't self-closing. This also handles the one-line `<Prompt …>text</Prompt>` shape directly.
- While inside a prompt, the close tag matches anywhere on the line, so children ending with `Copy this.</Prompt>` still exit the region.

Added six tests covering the regression shapes; the full suite (1425), typecheck, format check, and the per-file coverage gate all pass — `normalize.ts` is at 100% line / 99.7% branch. Also extended the changeset with a sentence on the anchored detection.

One inherited limitation left as-is: `</Prompt><Prompt>` on a single line only registers the close. It can't misplace a heading and no formatter emits that shape.

PR #67, opened 2026-07-16
Match prose bullets to counters (muted-foreground)

wadefletch:
Small theme fix: unordered-list bullets are noticeably lighter than the numbers in an ordered list, because the two markers pull from different tokens.

In `theme/entry.ts`:

```css
--tw-prose-counters: var(--blume-muted-foreground);
--tw-prose-bullets: var(--blume-border);   /* much lighter */
```

On the default light palette that's `oklch(0.552 …)` for numbers vs `oklch(0.92 …)` for bullets — the bullets nearly wash out on white. This points `--tw-prose-bullets` at `--blume-muted-foreground` too, so both markers match in light and dark.

Scoped to the one line; a patch changeset is included. Happy to adjust the token choice if you'd prefer bullets stay distinct — thanks for Blume! 🙏

### DIFF
--- a/.changeset/prose-bullets-muted-foreground.md
+++ b/.changeset/prose-bullets-muted-foreground.md
@@ -1,0 +1,5 @@
+---
+"blume": patch
+---
+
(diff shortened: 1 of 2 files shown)


haydenbleasel:
Great callout, thanks @wadefletch - aesthetically I prefer the current lighter style, however from an accessibility point of view this is likely to have an impact on people's ability to read, so I'll merge this 👍

PR #69, opened 2026-07-16
Themeable content-column width via `max-w-content`

wadefletch:
Draft / proposal — happy to adjust naming or approach.

The prose content column's max width is currently hardcoded as `max-w-[42rem]` in six places: the article, breadcrumb, mobile table of contents, page feedback, pagination, and the "last updated" line. Widening or narrowing the column means editing all six in lockstep, and a consumer can't retheme it at all.

This lifts that measure to a themeable token, following the same shape Blume already uses for radius (`--blume-radius` → `@theme --radius-blume` → `rounded-blume`):

- New `--blume-content-width: 42rem` base token.
- Mapped through `@theme inline { --container-content: var(--blume-content-width) }`, which Tailwind turns into a `max-w-content` utility.
- All six call sites use `max-w-content`.

The default is unchanged (`42rem`), so nothing moves out of the box. A consumer can now re-measure the whole column with a single override:

```css
:root { --blume-content-width: 48rem; }
```

**Verified** by building `apps/docs`: the generated CSS contains `--blume-content-width: 42rem` and `.max-w-content { max-width: var(--blume-content-width) }`, with no remaining hardcoded `max-width: 42rem`. A `minor` changeset is included.

A couple of open questions for you:
- Naming — `max-w-content` / `--blume-content-width` felt the most descriptive, but happy to rename (`--container-prose`, etc.).
- Whether you'd eventually want this surfaced through config (e.g. `theme.contentWidth`) the way `theme.radius` is — glad to follow up with that if so.

Thanks for Blume!

### DIFF
--- a/.changeset/themeable-content-width.md
+++ b/.changeset/themeable-content-width.md
@@ -1,5 +1,5 @@
 ---
-"blume": minor
+"blume": patch
 ---
 
 Add a themeable content-column width. A new `--blume-content-width` token (default `42rem`) is exposed as a `max-w-content` utility through Tailwind's `--container-content` theme key, and the article, breadcrumb, mobile table of contents, page feedback, pagination, and last-updated line now use it instead of hardcoding `42rem`. Override `--blume-content-width` to re-measure the whole column at once; the default is unchanged.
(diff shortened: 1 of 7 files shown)


haydenbleasel:
Great fix, highly appreciated 🙏

PR #71, opened 2026-07-16
Apply the deployment base to redirects instead of documenting around it

h-kono-it:
## Description

Add a note to the Redirects section: under `deployment.base`, a redirect's `to` needs the base written into it.

I hit this moving pages on a GitHub Pages project site. The build passed, but the emitted redirect pointed outside the base and 404'd — `from` had the base applied, `to` didn't. Writing the base into `to` fixes it, and that turns out to be intended: Astro reads the right-hand side as the final destination and [declined to change it](https://github.com/withastro/astro/issues/7774) ("we force usage of base in dev mode already I think it makes sense to also force you to use it in your redirects config").

So this is a docs gap rather than a bug, but I don't think I could have got there from the docs alone:

- **Redirects** shows `{ from: "/old", to: "/new" }` and doesn't mention base.
- **Subpath deploys** says of `deployment.base` that "internal links and assets are rewritten to include it" — a redirect target reads as an internal link.
- **Mount the docs under a path** says of `basePath`: "Write links as if mounted at root … Blume rewrites them, **along with redirects**" — accurate, since `applyBaseToRedirects` rewrites both sides.

I read that last sentence first and assumed it generalized. The failure is quiet: the build succeeds and the redirect just lands somewhere that doesn't exist.

Kept it prose-only to match the neighbouring note — happy to reshape it, and equally happy for you to close this if you'd rather word it differently or don't think it's worth the paragraph.

## Related Issues

N/A

## Checklist

- [x] I've reviewed my code
- [ ] I've written tests — n/a, docs only
- [ ] I've generated a change set file — n/a, `apps/docs` is private and in the changeset `ignore` list

### DIFF
--- a/.changeset/redirects-under-deployment-base.md
+++ b/.changeset/redirects-under-deployment-base.md
@@ -1,0 +1,7 @@
+---
+"blume": patch
+---
+
+Fix redirects escaping the site under `deployment.base`. A redirect's `to` was only ever rewritten with `basePath`, never with the deployment base, so with `base: "/docs"` a `to: "/new"` emitted a redirect to `/new` — outside the base, 404ing on a subpath deploy (GitHub Pages project sites, most commonly) with no build-time error. Astro applies `base` when it builds the match pattern for `from`, but resolves a destination either by regenerating it from a matching route's segments (which carry no base) or by passing it through verbatim — neither prepends `base`, so Blume now applies it to `to` itself. The two bases also compose correctly: `deployment.base` and `basePath` set together stack as `{base}/{basePath}`, which the old front-prepend could not produce in that order.
+
+The static host redirect files (`_redirects`, `vercel.json`, `blume-redirects.json`) had the mirror-image bug on the other side: they are matched against the real served URL, but `from` was written without the deployment base, so it never matched. Both sides now carry the full stack. Redirects are authored root-relative in every case, and a base already written into `to` by hand is preserved rather than doubled.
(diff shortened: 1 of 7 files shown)


haydenbleasel:
Thanks for this — the diagnosis was right, and the report was good enough to fix the cause rather than document it. I've pushed a commit that does that, so the note now says the opposite of what you wrote.

The reasoning: Blume already promises redirects are rewritten under `basePath`, and `deployment.base` is Blume's to apply too — so needing to hand-write it into `to` was the bug, not a gap in the docs. Astro treats the sides differently: it builds the `from` match pattern with `base` applied, but resolves a destination either by regenerating it from a matching route's segments (which carry no base) or by passing it through verbatim. Neither prepends `base`. That's also the mechanism behind your workaround — writing the base in stopped the destination matching a route, so it took the verbatim branch.

So `from` now takes `basePath` only, `to` takes the full `{base}/{basePath}` stack. Worth noting the old code couldn't produce that stack in the right order anyway: with both `deployment.base` and `basePath` set, no value of `to` was correct, since the base was always prepended at the front.

Your section also turned up a mirror-image bug next door. The static host files (`_redirects`, `vercel.json`, `blume-redirects.json`) are matched against the real served URL, but `from` was written without the deployment base — so it never matched. You wouldn't have seen it on GitHub Pages, which reads none of those. Both sides carry the full stack now.

Verified against a real static build with `base: "/docs"`: the emitted pages redirect to `/docs/introduction` and `/docs/nowhere-real` (covering both the route-matching and unmatched branches), and all three host files read `/docs/old /docs/introduction`.

Your `to: "/docs/new"` workaround keeps working — the rewrite is idempotent, so a hand-written base is preserved rather than doubled. Merging with your commit intact.

PR #74, opened 2026-07-16
Add search.popular for curated Cmd+K empty-state links

### LINKED ISSUE #73: Curate Cmd+K empty-state "Popular" links via config
## Problem

The search dialog's empty state ("Popular") is just the first six pages from the sidebar. On sites with a few top-level sections, that usually means whatever happens to sit first in the nav tree — not the pages people actually want to jump to from ⌘K.

Forking `Search.astro` to hardcode a list works, but it's a lot of surface area for a one-line product preference.

## Proposal

Add an optional `search.popular: { href, label }[]` (same shape as featured-style link lists elsewhere). When set and non-empty, use that list. When omitted/empty, keep today's sidebar fallback.

Internal routes stay base-less at resolve time so click-time `prefixBase` doesn't double-prefix under `deployment.base`.

Happy to open a PR if this sounds reasonable.

SutuSebastian:
## Summary

- Adds optional `search.popular: { href, label }[]` so sites can pin the ⌘K empty-state list instead of taking the first six sidebar pages.
- Empty/omitted keeps the existing sidebar fallback.
- Routes stay base-less until click time (same as sidebar-derived pages), so `deployment.base` doesn't get applied twice.
- Docs + changeset + tests.

Fixes #73.

## Test plan

- [x] `bun test` coverage for schema + `resolveSearchPopular` + runtime data
- [x] Pre-commit typecheck / ultracite / full package tests
- [ ] Smoke: set `search.popular` under a non-`/` `deployment.base` and confirm links resolve once
- [ ] Smoke: omit `popular` and confirm sidebar first-six still shows

### DIFF
--- a/.changeset/search-popular-pages.md
+++ b/.changeset/search-popular-pages.md
@@ -2,4 +2,4 @@
 "blume": patch
 ---
 
-Add `search.popular` to curate the Cmd+K empty-state link list. When set, each `{ href, label }` entry replaces the default first-six sidebar pages — useful on multi-tab sites where sidebar order surfaces the wrong section. Each `href` is authored root-relative and picks up `basePath` automatically (external URLs pass through). Omit or leave empty to keep the sidebar fallback.
+Add `search.popular` to curate the Cmd+K empty-state link list. When set, each `{ href, label, icon? }` entry replaces the default first-six sidebar pages — useful on multi-tab sites where sidebar order surfaces the wrong section. Each `href` is authored root-relative and picks up `basePath` automatically (external URLs pass through); `icon` takes a built-in icon name and defaults to a file glyph. Omit or leave empty to keep the sidebar fallback.
--- a/.changeset/search-popular-pages.md
+++ b/.changeset/search-popular-pages.md
@@ -1,0 +1,5 @@
+---
+"blume": patch
+---
+
+Add `search.popular` to curate the Cmd+K empty-state link list. When set, each `{ href, label }` entry replaces the default first-six sidebar pages — useful on multi-tab sites where sidebar order surfaces the wrong section. Omit or leave empty to keep the sidebar fallback.
--- a/.changeset/search-popular-pages.md
+++ b/.changeset/search-popular-pages.md
@@ -2,4 +2,4 @@
 "blume": patch
 ---
 
-Add `search.popular` to curate the Cmd+K empty-state link list. When set, each `{ href, label }` entry replaces the default first-six sidebar pages — useful on multi-tab sites where sidebar order surfaces the wrong section. Omit or leave empty to keep the sidebar fallback.
+Add `search.popular` to curate the Cmd+K empty-state link list. When set, each `{ href, label }` entry replaces the default first-six sidebar pages — useful on multi-tab sites where sidebar order surfaces the wrong section. Each `href` is authored root-relative and picks up `basePath` automatically (external URLs pass through). Omit or leave empty to keep the sidebar fallback.
--- a/apps/docs/content/docs/configuration/search.mdx
+++ b/apps/docs/content/docs/configuration/search.mdx
@@ -13,6 +13,22 @@ Open search with <Badge variant="accent">⌘K</Badge> (or `Ctrl K`), or press `/
 
 Queries match page **titles**, **descriptions**, and **body text**, with title matches ranked highest and descriptions above body.
 
+## Popular pages
+
+Before a reader types a query, the search dialog shows a **Popular** list. By default it is the first six sidebar pages — which on multi-tab sites often surfaces the wrong section. Pin the links you want instead:
+
+```ts blume.config.ts lineNumbers
+search: {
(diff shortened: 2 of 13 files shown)


haydenbleasel:
Thanks for this — the feature's shape is right and the docs/changeset/tests were all there. I've pushed two commits to your branch: `ea3eddfc` fixes a real bug, `0498d37c` adds the optional `icon`. Details below.

## `basePath` was never applied to curated hrefs

This is the substantive one. The PR body reasons carefully about `deployment.base` and gets it **right** — routes stay base-less and `prefixBase` applies the deployment base at click time. But `deployment.base` and `basePath` are two different bases, and only the first was handled.

`core/base-path.ts` spells out the model: `basePath` is a site-wide mount point applied **server-side at build time**, `deployment.base` is Astro's host-subdirectory base applied client-side, and "the two compose, stacking as `{deployment.base}/{basePath}/page`".

The convention for config-authored paths is stated at `core/navigation.ts:513`:

> Config-provided nav paths are authored as if mounted at root, so the base is applied here (idempotently, and only to internal paths — external URLs pass through). Content-derived sidebar routes are already based via `page.route`.

Every other config-authored path follows it — `featured.href`, tab paths, selector items. `search.popular` is config-authored and didn't. Meanwhile the sidebar fallback it replaces comes from `page.route`, which is already based via `sources/normalize.ts:471`.

Net effect: under `basePath: "/docs"`, a curated `/guides/start` rendered as `/guides/start` while the page is served at `/docs/guides/start` — **every internal popular link 404s**, and the two code paths disagree exactly when `basePath` is set.

The fix is small, since `withBasePath` is already idempotent and already no-ops on external URLs:

```ts
route: withBasePath(basePath, href)
```

Worth flagging: the test asserted the bug rather than catching it — it set `basePath: "/docs"` and expected `/guides/start` back, which is why CI was green. I split it into a no-`basePath` case (the branch it was actually testing) and a real `basePath` case covering the base being applied, a hand-written base not being doubled, and external/protocol-relative URLs passing through. Confirmed the new test fails against the old code.

## Redundant empty-list guards

The same "is it empty" question was asked three times: `buildRuntimeData` did `popular.length > 0 ? resolve(...) : []`, `Header.astro` mapped `[]` → `undefined`, and `Search.astro` checks `popularPages && popularPages.length > 0`. Since `resolveSearchPopular([])` is already `[]` and the Search check already handles `[]`, the first two were no-ops — dropped both. Only the `Search.astro` check earns its place (it also guards the slot-override path).

## Optional `icon` (`0498d37c`)

Added, since `navigation.featured` and `ai.ask.suggestions` (the closest analogue — also an empty-state list) both take one, and the rows hardcoded `svg("file")` — so the external "Blog" entry in the docs example rendered with a file icon.

The wrinkle is that the icon set is a server-only module. From `theme/chrome-icons.ts`:

> the full icon resolver (`./icons.ts`) pulls in library data far too large to ship to the browser, so it must never be imported from a client script

The island only has 5 hand-inlined chrome glyphs, so it can't resolve an arbitrary name. `AskAI.astro` already solved this for its suggestion icons — resolve to inline SVG in the Astro frontmatter, hand the island ready-to-render markup — so `Search.astro` now does the same. Rows without an icon, including every sidebar-derived fallback row, keep the file glyph, so existing behavior is unchanged.

One consequence worth knowing: because these render client-side, only **built-in names** work here — an image path or inline SVG (which `<Icon>` accepts elsewhere) falls back to the file glyph. That matches how `ai.ask.suggestions` already behaves for the same input, and it's documented.

Unknown names now warn at build time. `search.popular` sits outside the navigation model, so it misses the graph build's `validateNavIcons` — added `validateSearchPopularIcons`, called from `generateRuntime` where the search config is known, sharing the existing diagnostic logic. A typo'd icon otherwise just silently renders the default glyph, which is exactly what `nav-diagnostics.ts` exists to prevent:

```
[blume] WARN Unknown icon "not-a-real-icon" (popular link "TEMP Bad Icon") — it isn't in Blume's icon set. Use a built-in icon name.
```

## Docs

Added the `basePath` rule (write hrefs root-relative), the `icon` field and its built-in-names-only caveat, and a callout for an i18n sharp edge: `localizeRoute` prefixes routes per locale and Search gets the active locale's navigation, so the sidebar fallback follows the reader's language automatically — but a curated list is one global set shared by every locale. Not a blocker since the feature is opt-in, but it's a real cliff for the multi-section sites this is aimed at, so it's better stated than discovered.

## Verification

Full suite (1448 pass), typecheck, ultracite, and the per-file coverage gate (`popular.ts` and `nav-diagnostics.ts` both at 100%).

Since the icon path lives in `.astro` with no unit coverage, I also ran the docs site and drove the dialog: valid names render their glyph, an unknown name and an omitted icon both fall back to the file glyph, the external URL stays untouched, and the build warning above is real output from that run. The `basePath` fix is covered by unit tests at both the resolver and runtime-data layers rather than in the browser.

PR #75, opened 2026-07-16
Render authentication requirements in the native OpenAPI reference

marwanhilmi:
The renderer ignored security entirely, so authenticated endpoints were indistinguishable from public ones. Operations now show an Authorization section above their parameters — resolved from the operation's security (or the document root default) against components.securitySchemes — and the code samples send a matching placeholder credential.

OpenAPI semantics carry over as written: security: [] keeps an operation public, requirement entries render as "or" alternatives with the schemes inside one entry required together, an empty {} entry marks auth as optional, and OAuth scopes and scheme descriptions render per scheme. A requirement naming an unknown scheme is kept (labelled by its component name) instead of dropped. Samples take the first alternative; an explicit spec-declared header parameter still overrides the placeholder.

## Description

Please provide a brief description of the changes introduced in this pull request.

## Related Issues

Closes #<issue_number>

## Checklist

- [x] I've reviewed my code
- [x] I've written tests

Screenshot 2026-07-16 at 11 39 35 AM

## Additional Notes

You can see example from screenshot at test docs site [here](https://garage-docs-blume.grovemotorco.workers.dev/reference/operations/agents-claimview/). Please suggest any changes - happy to edit as needed.

### DIFF
--- a/.changeset/openapi-authorization.md
+++ b/.changeset/openapi-authorization.md
@@ -1,0 +1,5 @@
+---
+"blume": minor
+---
+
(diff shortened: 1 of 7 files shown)


haydenbleasel:
Fantastic work 🙏  Very much appreciate this

PR #81, opened 2026-07-17
Fix pnpm isolated linker dependency resolution

### LINKED ISSUE #72: Cannot resolve astro/config
Thanks for creating Blume — it’s a fantastic tool!

**Describe the bug**

When using pnpm's default `isolated` node linker, `blume dev` fails because the generated `.blume/astro.config.mjs` cannot resolve `astro/config`.

Astro is installed as a dependency of Blume inside pnpm's virtual store, but the `.blume/node_modules` dependency link that should expose Blume's dependencies is not created.

**Blume version**

I am using version `1.0.4`.

**To Reproduce**

Create a clean single-package pnpm workspace:

```sh
mkdir blume-pnpm-repro
cd blume-pnpm-repro

pnpm init

printf 'packages:\n  - "."\n' > pnpm-workspace.yaml

pnpm add blume@1.0.4
pnpm exec blume init --yes
pnpm run dev
```

The dev server fails with:

```text
ERROR Cannot find module 'astro/config' imported from '<project>/.blume/astro.config.mjs'
```

**Expected behavior**

`blume dev` should create the required dependency link for pnpm's isolated layout and start the Astro dev server.

**Root cause analysis**

pnpm's generated `node_modules/.bin/blume` wrapper adds Blume's dependency directories to `NODE_PATH`.

Blume's `ensureDepsLink()` checks whether Astro is already reachable from `.blume/` using `createRequire().resolve("astro/package.json")`.

Because CommonJS resolution honors `NODE_PATH`, this check successfully resolves Blume's Astro installation and returns early:

```ts
if (blumeAstro && outDirAstro === blumeAstro) {
  return null;
}
```

However, Vite subsequently loads `.blume/astro.config.mjs` through ESM resolution:

```js
import { defineConfig } from "astro/config";
```

ESM resolution does not use `NODE_PATH`, so this import fails.

I confirmed the difference in the same isolated installation:

```text
createRequire: <project>/node_modules/.pnpm/astro@.../node_modules/astro/package.json
ESM import: ERR_MODULE_NOT_FOUND
```

As a result, the preflight check considers Astro reachable, while the actual ESM runtime cannot resolve it. The intended `.blume/node_modules` link is therefore skipped.

This appears related to the isolated-linker support described in #8.

**Workaround**

Using a hoisted node_modules layout resolves the issue:

```ini
# .npmrc
node-linker=hoisted
```

After running `pnpm install` again, `blume dev` starts successfully.

**Environment**

- Command: `blume dev`
- Runtime: Node.js `24.18.0`
- Package manager: pnpm `10.34.4`
- Node linker: `isolated` (default)
- OS: macOS `26.5.2`, arm64
- Browser: N/A

deepso7:
## Description

Fix `blume dev` under pnpm's default isolated linker.

pnpm's generated CLI wrapper exposes Blume's dependencies through `NODE_PATH`, which CommonJS resolution honors but ESM resolution ignores. This caused Blume's dependency preflight to incorrectly conclude that `astro/config` was reachable from the generated `.blume` runtime.

The runtime now checks physical `node_modules` ancestors, matching the ESM resolution used by the generated Astro config. A regression test reproduces pnpm's `NODE_PATH` environment.

## Related Issues

Closes #72

## Checklist

- [x] I've reviewed my code
- [x] I've written tests

Not applicable — this change has no UI impact.

## Additional Notes

Verified end to end with a locally packed tarball under pnpm's default isolated linker. `blume dev` starts successfully and creates the expected `.blume/node_modules` link.

### DIFF
--- a/.changeset/fix-pnpm-node-path.md
+++ b/.changeset/fix-pnpm-node-path.md
@@ -1,0 +1,5 @@
+---
+"blume": patch
+---
+
+Fix `blume dev` under pnpm's default isolated linker. The generated runtime now checks Astro through the same physical `node_modules` ancestor lookup used by its ESM config, instead of mistaking pnpm's CommonJS-only `NODE_PATH` exposure for a resolvable `astro/config` import.
--- a/packages/blume/src/astro/generate.ts
+++ b/packages/blume/src/astro/generate.ts
@@ -149,20 +149,39 @@ const reactCompilerWarnings = (
       ]
     : [];
 
-/**
- * Realpath of the `astro` package node resolves from a directory, or null when
- * none resolves. Comparing this for `.blume/` against Blume's own deps tells
- * whether the runtime would bind to the *same* astro Blume uses or a different
- * one shadowing it (the hoisted-conflict failure mode).
- */
-const resolvedAstroPath = (fromDir: string): string | null => {
+/** Resolve Astro's package.json directly inside a node_modules directory. */
+const resolveAstroPackageJson = (modulesDir: string): string | null => {
   try {
-    const pkg = createRequire(
-      pathToFileURL(join(fromDir, "_.js")).href
-    ).resolve("astro/package.json");
-    return realpathSync(pkg);
+    return realpathSync(join(modulesDir, "astro", "package.json"));
   } catch {
     return null;
+  }
+};
+
+/**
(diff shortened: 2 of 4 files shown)


haydenbleasel:
Legend, highly appreciated @deepso7 🙏

Issue #78, opened 2026-07-17
Changelog timeline dates ignore the configured i18n locale (the last-updated stamp honors it) — intentional?


h-kono-it:
**Describe the bug**

With an i18n locale configured, the per-page "last updated" stamp is formatted in that locale, but the changelog timeline formats its dates with a hardcoded `"en"` — so one site shows dates in two languages.

The two formatters:

- `components/layout/RootLayout.astro` — `new Intl.DateTimeFormat(locale || "en", { dateStyle: "long", timeZone: "UTC" })` — honors the locale
- `astro/templates.ts` (the generated changelog page) — `new Intl.DateTimeFormat("en", { dateStyle: "long", timeZone: "UTC" })` — always English

**Blume version**

1.0.4 (the hardcoded `"en"` is still on `main`)

**To Reproduce**

1. Configure a single-locale Japanese site:
   ```ts
   i18n: { defaultLocale: "ja", locales: [{ code: "ja", label: "日本語" }] },
   lastModified: true,
   ```
2. Add one `type: changelog` entry with a `date`
3. `blume build`
4. A content page's footer shows `最終更新 2026年7月16日` (localized), while `/changelog` shows `July 15, 2026`

**Expected behavior**

I'd expect the timeline to follow the configured locale, matching the last-updated stamp — keeping `dateStyle: "long"`, so English sites see no change.

That said, if the hardcoded `"en"` is deliberate (changelogs often address a technical audience that reads English), that's a fair position too — I mainly wanted to flag that the two stamps currently disagree on one site. If you'd take a fix, happy to send a PR; for multi-locale sites I'd want your call on whether each localized changelog should use its own locale.

**Environment**

- Command: blume build
- Runtime: Node 22
- OS: macOS 26

haydenbleasel:
Good catch, and thanks for the thorough write-up — the repro made this quick to confirm. Not deliberate: the hardcoded `en` was written in the commit that added the changelog timeline, three days *before* i18n support landed. i18n made the last-updated stamp locale-aware and never revisited this template, so the two just drifted apart. No "changelogs are for English readers" intent behind it.

[...]

PR #89, opened 2026-07-19
Let served assets satisfy the audit's llms.txt stale-entry check

h-kono-it:
## Description

Thanks for 1.1.0! I upgraded right away — `seo.og.fonts` renders our Japanese OG cards perfectly (no more tofu, so I could finally drop the local font patch from #62), and the locale-aware changelog dates from #78 let me drop my other patch too. Zero patches now 🎉

While trying the new `blume audit` on the upgraded site, I hit what I think is a small false positive:

 ```
⚠ llms.txt lists a page the build does not serve
    https://h-kono-it.github.io/legal-check-helper/changelog/rss.xml  dist/llms.txt:99
```

but `dist/changelog/rss.xml` does exist in the build output. I think this is two 1.1.0 features meeting: llms.txt now includes RSS feed links, while `LLMS_TXT_STALE_ENTRY` resolves each entry against `context.byUrl`, which only contains HTML pages — so the feed link the generator itself emits is always flagged. Reproducible on any site with changelog entries and RSS enabled (both defaults).

This change also accepts any path the static file index serves, mirroring how `resolveRedirects` already composes pages + files into the served set.

Verified against a real site: on the same 1.1.0 build, stock `blume audit` flags `/changelog/rss.xml`; with this change the warning is gone and the rest of the report is unchanged.

## Related Issues

None — reporting and fixing in one go; happy to split out an issue if you prefer.

## Checklist

- [x] I've reviewed my code
- [x] I've written tests

### DIFF
--- a/.changeset/shiny-crabs-listen.md
+++ b/.changeset/shiny-crabs-listen.md
@@ -1,0 +1,5 @@
+---
+"blume": patch
+---
+
+Stop `blume audit` flagging the changelog RSS feed link in llms.txt as a stale entry. The stale-entry check compared each llms.txt target against built pages only, but the generator itself links non-page assets — the changelog RSS feed — so a target the static file index serves now counts as valid, the same way redirect targets may land on a served asset.
--- a/packages/blume/src/audit/checks/llms.ts
+++ b/packages/blume/src/audit/checks/llms.ts
@@ -91,7 +91,10 @@ export const llmsChecks: CheckModule = {
         continue;
       }
       listed.add(path);
-      if (!context.byUrl.has(path)) {
+      // A listed target may be a served asset rather than a page — Blume's own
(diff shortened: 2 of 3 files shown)


haydenbleasel (approved):
Thanks for this one — it's a model bug report *and* fix in a single PR. You traced the false positive exactly right: the generator links the RSS feed in llms.txt, but the stale-entry check only consulted the HTML page index, so the audit was flagging its own output. Composing pages + static files mirrors how `resolveRedirects` already builds the served set, the test is a faithful minimal repro, and the changeset is spot on.

Also really glad to hear `seo.og.fonts` and the locale-aware changelog dates let you drop both local patches — zero-patch upgrades are exactly what we're aiming for. 🎉

PR #96, opened 2026-07-19
feat(navigation): warn on duplicate sidebar order values

### LINKED ISSUE #94: blume doctor should detect duplicate sidebar order values
**Is your feature request related to a problem? Please describe.**

When two sibling sidebar nodes (pages or groups) end up with the same `sidebar.order` (or the same folder-`meta` `order`), the sidebar renders in a silently-arbitrary order with no warning at build time or from `blume doctor`.

`sortNodes` in `src/core/navigation.ts` breaks order ties by `label.localeCompare`:

```ts
// src/core/navigation.ts
const sortNodes = (nodes: MutableNode[]): void => {
  nodes.sort((a, b) => {
    if (a.order !== b.order) {
      return a.order - b.order;
    }
    return a.label.localeCompare(b.label);
  });
  ...
```

This is a reasonable fallback, but it means a real authoring mistake (two files both left at the default numeric-prefix order, or two folders both landing on the same manually-chosen `order`) never surfaces as feedback — it just quietly reorders itself alphabetically, which is rarely what the author intended.

While migrating a Starlight docs site to blume, we ended up with:

```ts
// content/meta-tags-properties/index.md frontmatter
sidebar: { order: 1 }
// content/meta-tags-properties/basic.md frontmatter
sidebar: { order: 1 }
```

Both landed on `order: 1`. The alphabetical tiebreak put `basic.md` ("Basic") *before* `index.md` ("MetaTags Properties") in the sidebar — i.e. the group's own landing page rendered second, after one of its children — which is inconsistent with every other group in the same site (where `index.md` is always first because it's the only one at that order). Nothing in `blume build` or `blume doctor` flagged this; it was caught only by manually diffing the rendered sidebar HTML against a sibling group.

We hit the same shape of bug twice more in the same migration for folder-level `order` in `meta.ts`/`meta.$.ts` (two groups both claiming `order: 4`), each requiring a follow-up fix-up commit once someone noticed the wrong position.

**Describe the solution you'd like**

Surface a diagnostic when two nodes in the same parent array (same scope: either siblings under one folder, or two root-level groups) share an `order` value, instead of only the label-based fallback. Something like:

```ts
// src/core/navigation.ts
const sortNodes = (nodes: MutableNode[], diagnostics?: Diagnostic[]): void => {
  const seenByOrder = new Map<number, MutableNode>();
  for (const node of nodes) {
    const prior = seenByOrder.get(node.order);
    if (prior && diagnostics) {
      diagnostics.push({
        code: "BLUME_DUPLICATE_SIDEBAR_ORDER",
        message: `"${prior.label}" and "${node.label}" both have sidebar order ${node.order}; falling back to alphabetical order.`,
        severity: "warning",
      });
    }
    seenByOrder.set(node.order, node);
  }
  nodes.sort((a, b) => (a.order !== b.order ? a.order - b.order : a.label.localeCompare(b.label)));
  for (const node of nodes) {
    if (node.kind === "group") sortNodes(node.children, diagnostics);
  }
};
```

`doctor.ts` already collects a `Diagnostic[]` from `project.diagnostics`, so this would slot into the existing `blume doctor` output rather than needing a new mechanism. `Number.NEGATIVE_INFINITY` for index pages and `Number.POSITIVE_INFINITY` for unprefixed non-index files would need excluding from the check (those ties are intentional/expected), but any two *explicit or numeric-prefix-derived* orders matching would be worth a warning.

Happy to put together a PR if this direction sounds right.

**Describe alternatives you've considered**

- Documenting "don't reuse order values" more prominently — doesn't help, since the collision is easy to introduce incrementally (e.g. inserting a new top-level page/group without checking every existing sibling's order first) and the site still builds and deploys fine, so nothing forces a re-read of the docs.
- A build-time hard error instead of a warning — considered, but a warning fits better since the alphabetical fallback is a legitimate (if surprising) resolution rather than a broken build; erroring could also be disruptive for large migrations with many pre-existing collisions to clean up incrementally.

oekazuma:
## Summary

Closes #94.

Two sibling pages or folders that resolve to the same explicit `sidebar.order`, numeric filename prefix, or folder-meta `order` currently fall back to a silent alphabetical tiebreak, with no feedback from `blume doctor`/`blume check`. This surfaces it as a `BLUME_DUPLICATE_SIDEBAR_ORDER` warning instead.

- `sortNodes` (`src/core/navigation.ts`) now collects ties at each sidebar level and reports one diagnostic per collision, naming every tied label.
- Nodes at the default fallback order (no numeric prefix, no explicit order — the common case) are excluded, so ordinary unordered folders don't get flagged.
- A node's order is also tracked as "authored" or not: an explicit `sidebar.order`, a numeric filename prefix, or a folder-meta `pages` rank counts as authored; a changelog entry's date-derived order does not. Two changelog entries published on the same day resolve to the identical order but aren't an authoring mistake, so they're excluded from the check the same way the default fallback is — only a tie where at least the pair itself is a deliberate authoring choice gets flagged.
- The diagnostics sink threads through `buildFileSystemSidebar` → `buildNavigation` (new optional `options.diagnostics`) → `buildContentGraph`, including the i18n per-locale build path, so a duplicate that only exists in one locale's tree is still caught.
- Added `BLUME_DUPLICATE_SIDEBAR_ORDER` to the diagnostic → docs-page map and to the Ordering section of the navigation docs.

Running `blume check` against this repo's own `apps/docs` site caught a real, pre-existing tie (two entries both left at `order: 3`) — included as a sanity check that the diagnostic behaves as intended on real content, not just fixtures. Left it as-is since fixing that content isn't in scope here.

## Test plan

- [x] `bun run typecheck`
- [x] `bun run check` (Ultracite lint + format)
- [x] `bun run test:coverage` (1770 pass, `navigation.ts` 99.55% function coverage)
- [x] New unit tests in `test/navigation.test.ts` (explicit page-order tie, folder-meta-order tie, default-order non-tie, no-sink-provided, same-day changelog entries not flagged, folder-meta `pages[]` rank colliding with an explicit `sidebar.order`) and `test/core.test.ts` (end-to-end through `buildContentGraph`)


haydenbleasel:
Thanks @oekazuma — this is a well-built feature, so rather than a round of review comments I pushed the polish directly (f44d9d9). What changed:

- **One warning per tie under i18n.** Untranslated pages are padded into every locale's tree from the fallback, so a tie in shared content was reported once per locale (an en/fr/de site printed it three times). `buildI18nNavigation` now collects each locale's diagnostics separately and dedupes on `code + file + message` — all locale-stable for padded pages, so a locale-specific tie in translated files still reports on its own.
- **Diagnostics now carry a `file`.** The warning anchors to the first tied page's source path, so `blume doctor`/`blume check` print a concrete location (folder-only ties have no single file, so `file` stays unset there). Running against this repo's docs now shows `at content/docs/03-faq.mdx` under the known `order: 3` tie.
- **Undated changelog entries are excluded.** A changelog file with no `date` frontmatter falls back to its numeric filename prefix, which for names like `2024-01-05-release.md` is a date stamp, not a rank — every such sibling tied at `2024` and warned. Their filename order is now treated as derived, matching your same-day-date exclusion.
- **Message formatting:** three-plus ties now read `"A", "B", and "C" all have…`.
- Replaced the `diagnosticsSink` helper with a default parameter on `buildFileSystemSidebar` — same effect, one less indirection. (Fun find: inlining `?? []` instead trips the complexity-20 lint on `buildNavigation`, which I suspect is how the helper came to be.)

Added tests for each: the per-locale dedup (i18n), file anchoring set/unset, the comma list, and the undated-changelog exclusion. Full suite and coverage gate are green. The `order: 3` tie your PR description caught in `apps/docs` is now fixed on main (819ac37f) — the FAQ page moved to the free slot 7 — so `blume check` on this repo comes up clean once this merges.

PR #97, opened 2026-07-19
feat(navigation): warn when an index page's title diverges from its folder's meta.title

### LINKED ISSUE #95: Group sidebar label and its index page's own title can silently fall out of sync across locales
**Is your feature request related to a problem? Please describe.**

A folder's sidebar group label and its own landing page's (`index.md`) page title/`<title>` are resolved from two completely independent sources, with no link between them — so a translator can update one and forget the other, and nothing surfaces the mismatch.

In `src/core/navigation.ts`:

- The **group label** comes from folder meta, applied in `applyFolderMeta`:
  ```ts
  const meta =
    folderMeta.get(metaKey(group.path, metaPrefix)) ??
    sharedMeta.get(group.path);
  if (meta) {
    group.label = meta.title ?? group.label;
    ...
  ```
  i.e. it's driven by that folder's `meta.ts` (or `meta.$.ts`) `title`.

- The **index page's own sidebar entry / page title**, since `index.md` is pushed into `parent.children` like any other page, comes from a totally different field:
  ```ts
  parent.children.push({
    ...
    label: page.meta.sidebar.label ?? page.title,
    ...
  ```
  i.e. it's driven by that page's own frontmatter `title` (or `sidebar.label`), independent of the folder's `meta.ts`.

Migrating an i18n docs site, we translated a group label via `meta.ts`:

```ts
// content/ja/meta-tags-properties/meta.ts
export default defineMeta({ title: "MetaTagsプロパティ", order: 10 });
```

...but left the group's own `content/ja/meta-tags-properties/index.md` frontmatter untranslated:

```md
---
title: MetaTags Properties
---
```

The sidebar renders correctly in Japanese ("MetaTagsプロパティ"), but the built page's own `<title>` tag (and, in some themes, its H1) render as the English "MetaTags Properties" — a silent, in-directory i18n regression that's easy to ship because the sidebar (the thing most people preview) looks completely correct. We only found it by grepping every `docs/content/**/index.md` for suspiciously-English titles.

**Describe the solution you'd like**

When an index page's own `title` is falling back to a default anyway (or, less invasively, when it's literally absent from frontmatter), let it inherit the enclosing folder's resolved `meta.title` instead of requiring the author to keep both in sync by hand. Something like, in `buildFileSystemSidebar` / around where `page.title` is read for an index page:

```ts
// pseudo-diff, buildFileSystemSidebar
const isIndex = isIndexStem(stem);
const folderTitle = isIndex ? resolvedFolderMeta(parent)?.title : undefined;
...
label: page.meta.sidebar.label ?? (isIndex ? folderTitle ?? page.title : page.title),
```

(The exact plumbing depends on whether folder meta for `parent` is available at this point in `buildFileSystemSidebar` or only after the later `applyFolderMeta` pass — might need reordering, or a second light pass over just index pages after `applyFolderMeta` runs.)

This wouldn't change anything for the common case (single-locale docs, or index pages that already set an explicit title matching the group), but it would remove the "two places to translate" trap for exactly the page most likely to be a plain landing page with no distinct title of its own.

**Describe alternatives you've considered**

- Have `blume doctor` flag index pages whose own title differs from a translated folder-meta title, as a locale-consistency hint, instead of changing the resolution/inheritance behavior itself. This is less invasive (no behavior change for existing sites) but only catches the problem after the fact rather than preventing it, and "differs from" is a fuzzy heuristic (a deliberately-different title is valid) versus inheritance being unambiguous when the field is simply absent.
- Leaving it purely as a documentation callout ("remember to translate both files") — doesn't hold up in practice, since this is exactly the kind of thing a docs contributor forgets under time pressure, and the site still builds/deploys with no error either way.

oekazuma:
## Summary

Closes #95.

A folder's sidebar group label and its own `index` page's frontmatter title are resolved from two completely independent sources (`applyFolderMeta`'s `meta.title` vs. `page.title`). Under i18n, a translator can update one and forget the other — the sidebar renders correctly while the built page's own `<title>` tag (and, in some themes, its H1) stays untranslated, with nothing surfacing the mismatch.

I considered the inheritance approach originally sketched in the issue (letting the index page's title fall back to the folder's resolved `meta.title`), but that changes `page.title` itself — which feeds the page's `<title>` tag, H1, breadcrumbs, and search index everywhere, not just the sidebar — for any site where an index page's title intentionally differs from its group label. That felt like too broad a behavior change to make silently. This PR instead goes with the alternative also noted in the issue: a `blume doctor`/`blume check` diagnostic (`BLUME_NAV_INDEX_TITLE_MISMATCH`) that surfaces the drift without changing existing resolution behavior for anyone.

- `buildFileSystemSidebar` (`src/core/navigation.ts`) now checks each index page against its containing folder's resolved `meta.title` (locale-aware, via the same `metaKey`/shared-meta lookup `applyFolderMeta` already uses) and reports a diagnostic when they diverge and the folder has an *explicit* title set (so ordinary folders with no `meta.ts` never trigger it).
- Only fires when the page itself has an *explicit* frontmatter `title` (`page.meta.title`). When it's absent, `page.title` falls back to the first heading or the filename — a value that almost never coincidentally matches a custom folder title, so comparing against it would flag noise on exactly the plain-landing-page case that's least worth warning about.
- The root group (an empty `group.path`) is skipped: a root-level `meta.ts`'s `title` is never converted into a rendered sidebar label (`toNavNode` only runs over `root.children`, never `root` itself), so a mismatch there wouldn't correspond to anything visible.
- Threaded through the same `options.diagnostics` sink pattern, from `buildFileSystemSidebar` → `buildNavigation` → `buildContentGraph`, including the per-locale i18n build path — so a locale-specific drift is caught for that locale specifically.
- Added `BLUME_NAV_INDEX_TITLE_MISMATCH` to the diagnostic → docs-page map and to the navigation docs.

Running `blume check` against this repo's own `apps/docs` site surfaced two real, pre-existing divergences (`Configuration file` vs. `Configuration`, `Pages` vs. `Content`) — included as a sanity check that the diagnostic fires on real content. Left that content as-is since it may be intentional; the diagnostic is a warning, not an error.

## Test plan

- [x] `bun run typecheck`
- [x] `bun run check` (Ultracite lint + format)
- [x] `bun run test:coverage` (1770 pass, `navigation.ts` 99.55% function coverage)
- [x] New unit tests in `test/navigation.test.ts` (mismatch, matching title, no folder meta, non-index page ignored, no explicit frontmatter title, root group skipped) and `test/core.test.ts` (end-to-end through `buildContentGraph`)


haydenbleasel:
Thanks for this — the diagnostic approach over silent title inheritance was the right call, and the meta lookup mirroring `applyFolderMeta` exactly means it always compares against the label the sidebar actually renders. I pushed c4a4c84 with three refinements found while reviewing the i18n paths:

**1. One report per mismatch, not one per locale.** Every locale's navigation build shares the same diagnostics sink, and each locale's page set includes fallback-filled entries from the default locale. Under the `dot` parser (where `metaPrefix` is empty for every locale) or with shared `meta.$` files, a single divergence therefore repeated once per configured locale. `localePagesFor` now stamps `fallback: true` on filled entries (new optional field on `PageRecord`), and the diagnostic skips them — each page is only checked in the locale that owns it, so a mismatch can't repeat and no dedup pass is needed.

**2. No more misleading suggestion on untranslated pages.** Previously, a translated `fr/guides/meta.ts` with a not-yet-translated `guides/index.mdx` flagged the *English* source file and suggested setting its frontmatter title to the French meta title — which would break the default locale. That case is the same fallback fill, so it's covered by the exemption above; the default locale's own build still checks the real page.

**3. Sidebar-hidden index pages are now checked.** The check ran after the `sidebar.hidden` early-continue, but a hidden index page still renders its own `<title>` — exactly the drift this diagnostic exists for. It now runs before the filter, deriving the folder path from `navPath` instead of the tree node (hidden pages still stay out of the sidebar, tested).

Also removed the `diagnosticsSink` helper — it turned out to exist to stay under the complexity-20 lint cap on `buildNavigation`, so the sink is passed through as `Diagnostic[] | undefined` and guarded at the push site instead. Docs and the changeset mention the fallback exemption.

Added five tests: fallback exemption and hidden-index at the unit level, plus three integration tests in `i18n.test.ts` (dot-parser mismatch reported exactly once; translated-meta-with-untranslated-page stays quiet; a genuinely drifted French index page is flagged with the French file in the message). All five fail against the pre-fix source. Full suite: 1775 pass, coverage gate green.

PR #100, opened 2026-07-19
fix(components): size Component preview panes to the rendered example

### LINKED ISSUE #98: Component preview panes size to the source line count, so real components clip or float in dead space
I spent today migrating Neon UI's docs (a shadcn registry, ~28 component demos) from a hand-rolled preview pipeline to `<Component>`. Almost everything worked: the glob source pointing outside the project root, variant demos, `examples.css` with an extra `@source` for our registry package. Sizing is the one thing that made me revert.

The pane height is derived from the code, not from what the code renders:

```js
// Component.astro
const LINE_PX = 21;
const PADDING_PX = 36;
const paneHeight = Math.min(400, Math.max(288, lineCount * LINE_PX + PADDING_PX));
const paneStyle = `height:${paneHeight}px`;
```

Line count and rendered height only correlate for small examples. For real components they diverge hard in both directions. Our chat demo is about 20 lines of source but renders a 520px conversation UI, so it clips at the 400px ceiling, cutting the input row in half:

Image

Our status badge demo goes the other way: the source is long enough to earn a 400px pane, and the rendered output is four rows of badges floating in it:

Image

You can see a mild version of the same thing on useblume.dev, where the counter example is a single button sitting in a 393px pane:

Image

There's currently no way around it from userland. `Props` is just `{ path }`, the height is an inline style, and nothing in the frame reports its size back.

What I'd actually want is the iframe sizing to its content: the frame page observes its own height (ResizeObserver, posted via postMessage) and the parent sets the iframe height from that, ideally with a min and a transition so hydration doesn't jump. That would fix every case without any API surface.

If that's more than you want in there, a `height` prop on `<Component>` that overrides the computed pane height would unblock registries like ours, and the line-count formula stays the default.

Happy to PR either one, just tell me which direction you'd take.

v blume 1.1.1, and I checked the same code is on main as of today.

jal-co:
## Description

Follows the direction from the issue thread: the frame reports its rendered height, no new API surface.

- The generated frame page wraps the example in a marked `div` and watches it with a `ResizeObserver`, posting `{ type: "blume:example-height", height }` to the parent. The report includes the body's 1.5rem padding so the parent applies the number as-is, and the target origin is pinned rather than `"*"`. Direct opens (no parent) skip out, same as the theme sync.
- `Component.astro` registers one `message` listener for the whole page, matches the sender to its iframe via `event.source`, and applies the measured height to **both** tab panels, so the shared-height invariant holds and toggling Preview/Code never shifts the layout.
- The line-count formula stays as the SSR/no-JS initial height with the 288px floor. The 400px ceiling is dropped — the code tab already scrolls inside whatever height it's given (`pre.blume-source`). A `motion-safe` height transition smooths the settle when the lazy iframe loads.
- Since the observer keeps reporting, examples that resize after load keep the pane in sync.

## Related Issues

Closes #98

## Checklist

- [x] I've reviewed my code
- [x] I've written tests

Smoke-tested against the built docs site (headless Chromium over `dist/`): the counter example's panes were 393px from the line count; at runtime the frame reported 86px of content and both panels settled at the 288px floor through the message path:

```json
{
  "panelHeights": ["288px", "288px"],
  "exampleHeight": 86,
  "iframeBox": 288
}
```

## Additional Notes

Ran `bun run check`, `bun run typecheck`, and `bun run test` (all green), plus a full `apps/docs` build to confirm the generated frame pages and the inlined listener land in the output. One test was added to `templates.test.ts` locking in the wrapper marker, the observer, the message type, and the pinned origin.


packages/blume/src/components/content/Component.astro
```diff
@@ -81,3 +92,39 @@ const paneStyle = `height:${paneHeight}px`;
...
+      "iframe[data-blume-example-frame]"
+    );
+    const frame = Array.from(frames).find(
+      (candidate) => candidate.contentWindow === event.source
+    );
+    const tabs = frame?.closest("blume-tabs");
+    if (!tabs) {
+      return;
+    }
+    const height = `${Math.max(MIN_PANE_PX, event.data.height)}px`;
```
haydenbleasel: **Feedback loop for viewport-height examples.** An example styled with `h-screen`/`100svh` sizes itself to the frame's viewport — which is exactly the pane height this listener sets. Each report then comes back as `frameHeight + 48`, the pane grows by 48px, the example follows, the observer fires again: unbounded growth, smoothly animated. The old 400px ceiling was accidentally containing this case; with it gone there's no backstop.

A cheap guard here: ignore a report within ~48px of the frame's current `offsetHeight` (an echo can never differ by more than the padding, a real content change almost always does), or refuse growth beyond something generous like `window.innerHeight`. Either keeps the fix intact for the chat-demo case from #98.

packages/blume/src/components/content/Component.astro
```diff
@@ -47,30 +47,41 @@ const codeHtml = entry
...
+// line-count estimate (≈21px/line + padding, 18rem floor) is only the initial
+// SSR/no-JS height: once the frame loads it reports its rendered height (see
+// the script below) and both panes follow it, so previews fit the example —
+// including ones that grow or shrink after load. The code tab scrolls inside
+// whatever height it's given (`pre.blume-source`), so no ceiling is needed.
 const LINE_PX = 21;
 const PADDING_PX = 36;
 const lineCount = entry ? entry.code.replace(/\n+$/u, "").split("\n").length : 0;
-const paneHeight = Math.min(400, Math.max(288, lineCount * LINE_PX + PADDING_PX));
+const paneHeight = Math.max(288, lineCount * LINE_PX + PADDING_PX);
```
haydenbleasel: Dropping the ceiling on the *measured* height is the fix, but this is only the placeholder — and uncapped, a 200-line source now renders a ~4200px pane that collapses to the measured height when the lazy iframe loads mid-scroll. The 200ms transition can't hide a shift that size.

Consider keeping a ceiling on the estimate alone (400–600px). That bounds the settle to the clipping cases from the issue (e.g. 400 → 520px) instead of thousands of pixels. No-JS readers aren't hurt: without JS the tabs never build triggers anyway, and `pre.blume-source > code` scrolls at any height.

packages/blume/src/astro/templates.ts
```diff
@@ -2020,7 +2026,37 @@ const Example = entry.Component;
...
+        if (window.parent === window) {
+          return;
+        }
+        const wrapper = document.querySelector("[data-blume-example]");
+        if (!wrapper) {
+          return;
+        }
+        // The body's 1.5rem padding frames the example; fold it into the
+        // report so the parent can apply the number as-is.
+        const PADDING_PX = 48;
```
haydenbleasel: This hardcodes the body's `1.5rem` padding as 48px, which assumes a 16px root font size — but the user's `examples.css` is injected after Blume's defaults precisely so their tokens win, and a root `font-size` override would skew every report. Reading it once from the live value removes the assumption:

```js
const bodyStyle = getComputedStyle(document.body);
const padding =
  parseFloat(bodyStyle.paddingTop) + parseFloat(bodyStyle.paddingBottom);
```

packages/blume/src/components/content/Component.astro
```diff
@@ -81,3 +92,39 @@ const paneStyle = `height:${paneHeight}px`;
...
+  // measured height replaces the server's line-count estimate on both tab
+  // panels together, preserving the shared-height invariant that keeps
+  // Preview/Code toggles from shifting the layout.
+  const MIN_PANE_PX = 288;
+
+  window.addEventListener("message", (event) => {
+    if (
+      event.origin !== window.location.origin ||
+      event.data?.type !== "blume:example-height" ||
+      typeof event.data.height !== "number"
```
haydenbleasel: Nit: `typeof … === "number"` admits `NaN` and `Infinity` — `Math.max(288, NaN)` yields `"NaNpx"`, which the browser silently drops (harmless fallback to the estimate, but accidental). `!Number.isFinite(event.data.height)` is the intended check at the same cost.

haydenbleasel (changes requested):
Thanks — this lands exactly the direction from #98, and the details are largely right: origin pinned on both ends, sender matched via `event.source` rather than trusting the payload, the wrapper observed instead of the body (with the comment explaining why), and the shared-height invariant preserved across both panels. I verified the hooks the listener relies on (`blume-tabs`, `[data-blume-tab-panel]`, and `pre.blume-source > code { overflow: auto }` backing the no-ceiling claim), and the changeset/docs/test all follow the house conventions.

One real issue before merge: the measurement loop has no guard against examples whose height tracks the frame's viewport (`h-screen`/`100svh`), which turns the report cycle into unbounded growth — the old 400px ceiling was accidentally containing that case. Details inline, plus three smaller suggestions (capping the initial estimate to bound the settle shift, the 16px-rem assumption in `PADDING_PX`, and a `Number.isFinite` check).

packages/blume/src/components/content/Component.astro
jal-co: Good catch, and you're right that the old ceiling was doing accidental containment duty. Went with the viewport clamp: the applied height is now capped at window.innerHeight, so the echo cycle parks once the pane reaches the viewport and the observer goes quiet. Picked it over the ~48px echo guard because the guard's tolerance would have had to track the body padding, which is now read from computed style per your other comment, and a fixed tolerance would also swallow legitimate sub-48px content changes (an accordion opening by 30px, say). The chat-demo case from #98 still works; anything genuinely taller than the screen scrolls inside the frame, which a taller-than-screen pane wouldn't have saved it from anyway. Fixed in 8513af7.

packages/blume/src/components/content/Component.astro
jal-co: Agreed, the uncapped placeholder was trading one layout shift for a much bigger one. The estimate now ceilings at 400px (matching the old value) while the measured height stays uncapped below the viewport, so the settle is bounded to the clipping cases from the issue. Comment in the file spells out the asymmetry so the two limits don't get conflated later. Fixed in 8513af7.

packages/blume/src/astro/templates.ts
jal-co: Took your snippet nearly verbatim. The frame now reads paddingTop + paddingBottom from getComputedStyle(document.body) once and folds that into the report, so a root font-size override in examples.css flows through instead of being silently off by the rem delta. Added a test asserting the template reads the computed style and no longer contains the hardcoded 48. Fixed in 8513af7.

packages/blume/src/components/content/Component.astro
jal-co: Yep, that was sloppy. Number.isFinite it is, so NaN and Infinity payloads are rejected up front instead of riding the accidental NaNpx fallback. Fixed in 8513af7.

haydenbleasel:
Pushed bdfe614a with three small follow-ups from review — no behavior changes to the measurement design itself:

- **Single source of truth for the 288px floor.** It was duplicated as a literal in the frontmatter estimate and as `MIN_PANE_PX` in the client script; if one changed without the other, panes would visibly jump on load. The constant now lives only in the frontmatter and rides to the client on the iframe's `data-blume-min-pane` attribute, which the listener reads when clamping.
- **Re-clamp on window resize.** A pane capped by a small viewport used to stay small after the window grew, because the frame's content had stopped changing size so its ResizeObserver had nothing new to report. The listener now stores each frame's last raw report in `data-blume-reported-height` and reapplies the clamp on `resize`, so the pane recovers (and shrinks promptly when the window shrinks). The h-screen convergence argument is unchanged — the clamp math is the same, just recomputed on one more trigger.
- **Dropped a no-op test assertion.** `expect(out).not.toContain("PADDING_PX = 48")` asserted the absence of a string that never existed in the template (`PADDING_PX` is a `Component.astro` constant, and it's 36); the positive `getComputedStyle(document.body)` assertion on the line above already locks in the live-value behavior.

`bun run check`, `bun run typecheck` (including `astro check` over the docs), and `bun run test` (1807 pass) are all green on the branch.

PR #101, opened 2026-07-19
Support custom syntax highlighting themes

### LINKED ISSUE #76: Support custom syntax highlight themes
The general theme is generally customizable so adding support for custom syntax highlighting themes would be great. 

Something like this:
```
import myTheme from "./themes/acme-dark.json" with { type: "json" }; // => Shiki compatible json

export default defineConfig({
  markdown: {
    codeBlocks: {
      theme: { light: "github-light", dark: myTheme },
    },
  },
});
```

Have this implemented [here](https://github.com/marwanhilmi/blume/commit/6db1f3ce3f913f3c4e0998af09ae01355814d26a) and can submit PR.

marwanhilmi:
## Description

This adds support for custom Shiki syntax highlighting themes. The `codeThemeSchema` schema could probably be improved but this should help for now.

## Related Issues

Closes #76 

## Checklist

- [x] I've reviewed my code
- [x] I've written tests

### DIFF
--- a/packages/blume/src/components/content/diff.ts
+++ b/packages/blume/src/components/content/diff.ts
@@ -8,13 +8,15 @@
  * a unified patch (string or `.patch`/`.diff` file), a pair of file paths, or a
  * pair of inline strings.
  */
+import { createHash } from "node:crypto";
 import { readFile } from "node:fs/promises";
 
+import { registerCustomTheme } from "@pierre/diffs";
 import { preloadDiffHTML, preloadPatchDiff } from "@pierre/diffs/ssr";
 import { isAbsolute, join } from "pathe";
 
 import { DEFAULT_CODE_THEMES } from "../../markdown/themes.ts";
-import type { CodeThemes } from "../../markdown/themes.ts";
+import type { CodeTheme, CodeThemes } from "../../markdown/themes.ts";
 
 export interface DiffOptions {
   /** Path to the "after" file, resolved relative to {@link DiffOptions.root}. */
@@ -46,6 +48,39 @@ const resolvePath = (path: string, root: string): string =>
 const readText = (path: string, root: string): Promise<string> =>
   readFile(resolvePath(path, root), "utf-8");
 
+const registeredDiffThemes = new WeakMap<object, string>();
+
+/**
+ * Pierre accepts custom Shiki themes through its registry, while Shiki itself
+ * accepts the object directly. Register each configured object under a name
+ * derived from its content so registration survives dev-server reloads: the
+ * same theme re-registers under the same name (harmless overwrite), and an
+ * edited theme gets a fresh name instead of a stale cached entry.
+ */
+const diffThemeName = (theme: CodeTheme, mode: "dark" | "light"): string => {
+  if (typeof theme === "string") {
+    return theme;
+  }
+  const cached = registeredDiffThemes.get(theme);
+  if (cached) {
+    return cached;
+  }
+  const hash = createHash("sha256")
+    .update(JSON.stringify(theme))
+    .digest("hex")
+    .slice(0, 12);
+  const name = `blume-custom-${mode}-${hash}`;
+  const registered = { ...theme, name, type: theme.type ?? mode };
+  registerCustomTheme(name, () => Promise.resolve(registered));
+  registeredDiffThemes.set(theme, name);
+  return name;
+};
+
+const diffThemes = (themes: CodeThemes): { dark: string; light: string } => ({
+  dark: diffThemeName(themes.dark, "dark"),
+  light: diffThemeName(themes.light, "light"),
+});
+
 /**
  * Resolve `<Diff>` inputs to a prerendered HTML string. Throws when no input
  * group is supplied or a pair is half-specified, so the component can degrade
@@ -67,7 +102,7 @@ export const renderDiff = async (options: DiffOptions): Promise<string> => {
   if (patch !== undefined || src !== undefined) {
     const text = patch ?? (await readText(src as string, root));
     const result = await preloadPatchDiff({
-      options: { theme },
+      options: { theme: diffThemes(theme) },
       patch: text,
     });
     return result.prerenderedHTML;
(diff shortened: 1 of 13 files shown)


packages/blume/src/core/schema.ts
```diff
@@ -950,9 +951,25 @@ const githubConfigSchema = z.strictObject({
...
+    return false;
+  }
+  const theme = value as Record<string, unknown>;
+  return (
+    typeof theme.colors === "object" &&
+    theme.colors !== null &&
+    !Array.isArray(theme.colors) &&
+    Array.isArray(theme.tokenColors)
+  );
+}, "Expected a Shiki theme name or custom theme object");
```
haydenbleasel, line 981: This schema is stricter than the `CodeTheme` type it validates, so some themes typecheck in `blume.config.ts` and then fail at build time.

Shiki's `ThemeRegistrationAny` makes every field optional and also accepts the TextMate raw form, which uses `settings` instead of `tokenColors` — for example, a theme built with Shiki's `createCssVariablesTheme()`, or older theme JSONs, have `settings` and often no `colors`. Those are valid `CodeTheme` values but get rejected here with "Expected a Shiki theme name or custom theme object", which is confusing DX given the type said it was fine.

Suggested loosening — accept either token list, and only shape-check `colors` when present:

```ts
return (
  (theme.colors === undefined ||
    (typeof theme.colors === "object" &&
      theme.colors !== null &&
      !Array.isArray(theme.colors))) &&
  (Array.isArray(theme.tokenColors) || Array.isArray(theme.settings))
);
```

packages/blume/src/components/content/diff.ts
```diff
@@ -46,6 +47,35 @@ const resolvePath = (path: string, root: string): string =>
...
+  const cached = registeredDiffThemes.get(theme);
+  if (cached) {
+    return cached;
+  }
+  const name = `blume-custom-${mode}-${nextDiffThemeId}`;
+  nextDiffThemeId += 1;
+  const registered = { ...theme, name, type: theme.type ?? mode };
+  registerCustomTheme(name, () => Promise.resolve(registered));
+  registeredDiffThemes.set(theme, name);
+  return name;
```
haydenbleasel: Two edges in this registration scheme:

1. **Stale themes across dev-server reloads.** The `WeakMap` and counter live in this module, but Pierre's theme registry lives in `@pierre/diffs` (a render-external dep cached in the Node process). If this module re-executes while Pierre stays cached — e.g. a config edit in dev — the counter resets to 0, the re-register of `blume-custom-dark-0` hits Pierre's `DuplicateThemeError` path, which logs a `console.error` and **keeps the old loader**. So after editing a custom theme, `<Diff>` keeps rendering the stale one until a full restart, with console noise on every reload.

2. **The cache ignores `mode`.** Passing the same theme object to both `dark` and `light` returns the name registered first (with `type: "dark"`), so light mode gets a dark-typed theme when the object has no explicit `type`.

Deriving the name from a content hash of the theme (e.g. `blume-custom-${hash(theme)}-${theme.type ?? mode}`) fixes both: identical content collides harmlessly on re-registration, changed content gets a fresh name and registers cleanly, and the two modes stop sharing a cache entry.

haydenbleasel (commented):
Thanks for this — the overall shape is right: theme objects pass through to Shiki untouched, and registration only happens where Pierre demands names. I verified the branch locally: all 1810 tests pass, the per-file 98% coverage gate passes (`diff.ts` at 100% functions / 98.61% lines), `tsgo` typecheck passes, and `ultracite check` is clean. I also confirmed the `@pierre/diffs` root entry and the `/ssr` entry resolve to the same internal `themeResolver` singleton, so registering via the root import and rendering via `/ssr` is sound — and the diff test proves it end-to-end.

Two things worth addressing before merge — see the inline comments:

1. **Schema is stricter than the `CodeTheme` type** — valid Shiki themes using the TextMate `settings` form (e.g. `createCssVariablesTheme()`) typecheck but fail validation.
2. **Diff theme registration can go stale across dev-server reloads** — a content-hash-derived name would fix it.

Minor notes, no action needed:
- `z.custom` still accepts any string, so a typo'd bundled name fails at Shiki load rather than validation — same as before this PR, no regression.
- Custom theme objects get `JSON.stringify`'d into the generated Astro config and `data.json`, so full VS Code themes add some weight to generated files. Server-side only, harmless.
- Docs additions (import attribute for JSON, mixing bundled + custom) read well and match the implementation.

packages/blume/src/components/content/diff.ts
marwanhilmi: Updated to generate a hash and include mode in theme name.

packages/blume/src/core/schema.ts
marwanhilmi: Updated to the suggestion and added some additional tests for different theme data.

haydenbleasel:
Thanks for this — the Pierre adapter design (content-hashed names, per-type memo) is solid, and I've verified the docs' `with { type: "json" }` example loads correctly through jiti, which is what actually loads `blume.config.ts`. CI was green, so I pushed a small follow-up commit (7790ca2b) addressing three minor things rather than round-tripping review comments:

**1. Guarded duplicate theme registration.** Pierre's `registerCustomTheme` catches its own `DuplicateThemeError` but logs `SharedHighlight.registerCustomTheme: theme name already registered` via `console.error`. The `WeakMap` memo only dedupes per object *identity*, so a second identical-content object — a dev-server config reload, or the same theme JSON imported in two places — would spam that error on every `<Diff>` render. There's now a module-level `Set` of registered names that skips the re-register (which is safe precisely because the name is content-derived: same name ⇒ same theme).

**2. Schema now accepts colors-only themes.** `codeThemeSchema` required `settings` or `tokenColors` to be an array, which rejected a minimal-but-valid VS Code theme like `{ name, type, colors: { "editor.background": "#000" } }` even though Shiki renders it fine from editor fg/bg. The validator now requires each *present* field to have the right shape and at least one of the three to be present — so `{}` and malformed fields (`settings: {}`, `colors: []`) are still rejected with the same error message and path.

**3. Test coverage for the inline-code path.** `inline-code.ts` only had its type widened, so the inline `` `code`{:lang} `` path was the one surface without an end-to-end custom-theme test. Added one, plus a colors-only rendering test through `highlightCode` to back the schema change.

Also downgraded the changeset from `minor` to `patch` to match how we're versioning releases.

Full suite (1817 tests + per-file coverage gate), typecheck, and formatting all pass locally with the follow-up.

PR #115, opened 2026-07-23
Pre-bundle epub-gen-memory so the EPUB export works in dev

h-kono-it:
## Description

The **Export to EPUB** page action throws in dev and never produces a file:

```
[blume] EPUB export failed TypeError: epub is not a function
```

Production builds are fine — this is dev-only.

### Cause

`epub-gen-memory`'s browser bundle is a browserified UMD, and Blume's `import("epub-gen-memory/bundle")` lives inside `node_modules/blume`, which Vite's optimizer scan doesn't crawl. So it's never discovered, never pre-bundled, and in dev it's served as raw ESM — where the UMD finds neither `exports` nor `define`, falls through to its global branch, and strands its callable on `window.epubGen` while exporting nothing. `epubModule.default` is `undefined`, so the `.default?.default ?? .default` unwrap can't recover it.

This is the same failure mermaid already has an `optimizeDeps.include` entry for, and the comment there notes it only affects dev.

One wrinkle: `"blume > epub-gen-memory"` alone doesn't fix it. The import targets the `/bundle` subpath, which is a separate entry from the package root, so the include has to name it — the same "subpath needs naming explicitly" point the `takumi-js/helpers` note in this file already makes for `ssr.external`.

### Verification

Measured on a real site (GitHub Pages project site, `deployment.base` set) at `blume@1.1.2`; the relevant code is unchanged on `main`. Then re-checked end to end by patching the generated config's source and letting `.blume` regenerate from scratch:

| | dev before | dev after | build |
| --- | --- | --- | --- |
| resolves to | raw `bundle.min.js` | `.vite/deps/…_bundle.js` | bundled |
| error | `epub is not a function` | none | none |
| EPUB produced | no | yes | yes (5,780 B) |
| `window.epubGen` | `object` (stranded) | `undefined` | `undefined` |

## Related Issues

None — reporting and fixing together, since the cause matched an existing in-repo pattern. Happy to split out an issue if you'd prefer.

## Checklist

- [x] I've reviewed my code
- [x] I've written tests

### DIFF
--- a/.changeset/olive-pugs-repeat.md
+++ b/.changeset/olive-pugs-repeat.md
@@ -1,0 +1,5 @@
+---
+"blume": patch
+---
+
+Fix the EPUB page action failing in dev with `epub is not a function`. `epub-gen-memory`'s browser bundle is a browserified UMD, and its dynamic import lives inside `node_modules/blume`, which Vite's optimizer scan doesn't crawl — so in dev it was served as raw ESM, where the UMD finds no `exports`/`define`, exposes no `default`, and strands its callable on `window.epubGen`. It now joins mermaid in `optimizeDeps.include`, naming the `/bundle` subpath that is actually imported, since optimizing the package root leaves that entry unoptimized. Production builds already bundled it correctly and are unchanged.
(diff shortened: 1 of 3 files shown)


haydenbleasel:
Nice one, great fix 🚀  Appreciate it!

Issue #121, opened 2026-07-26
configured-integrations e2e times out on macOS — the spawned dev process never listens


h-kono-it:
**Describe the bug**

`packages/blume/test/configured-integrations.test.ts` (added in #83) times out consistently on my
machine, which makes the `pre-commit` hook unusable since it runs `test:coverage`. I'm not sure
whether this is specific to my setup, so I wanted to share what I observed in case it helps.

The test fails at its 300s budget with no inner error — none of the 30s `waitUntil` timeouts fire,
and the retry in `startDevReady` never kicks in.

```
(fail) runs configured integrations in order for build and dev, regenerates once on edit, and applies edits after process restart [300001.08ms]
  ^ this test timed out after 300000ms.

 1 pass
 1 fail
```

**What I observed**

Sampling the process tree every 30s while the test ran:

```
13:22:48  bun … blume.mjs build --isolated          (progressing)
13:23:18  bun … blume.mjs dev --port 62993   PPID=1  CPU 0.0%
13:23:48  (same pid)                         PPID=1  CPU 0.0%
   …
13:27:19  (same pid, 4m31s)                  PPID=1  CPU 0.0%
```

The dev process is spawned and reaches `astro:server:setup` — the fixture's marker log holds
`config:first`, `config:second`, `server:first`, `server:second` — then sits at 0% CPU and never
listens. It is also reparented to PID 1.

At the same time the test's own poll loop looks stalled: `waitForDevServer` should throw after 30s
and `startDevReady` should relaunch once, but neither happens. So both sides appear stuck, not just
the child.

**What I ruled out**

- **Astro dev itself.** I recreated the fixture by hand (same `blume.config.ts`, same
  `node_modules/blume` symlink into the package root) and ran
  `bun bin/blume.mjs dev --host 127.0.0.1 --port 4399` directly — ready in 2s, HTTP 200, all four
  markers written.
- **`blume build --isolated` being slow.** Run by hand in the same fixture: 3.3s, exit 0.
- **Stale `blume-integrations-*` fixtures** from killed runs: the leftovers were created *after* the
  first failure, so they weren't the cause.
- **`bun` version.** 1.3.14, the current release.
- **`dist/cli/index.js` vs the `src/cli/index.ts` fallback** in `bin/blume.mjs`: fails either way.
- **File descriptor limits.** `ulimit -n` is 1048576.

So the difference seems to be running the CLI through `Bun.spawn` with `stdout`/`stderr` set to
[...]

PR #122, opened 2026-07-26
PR: Let a header tab declare its link target with `href`

h-kono-it:
## Description

A tab's `path` scopes its sidebar section and doubles as its link target. When that `path` isn't a page of its own (a folder with no `index.mdx`), linking straight to it would 404, so `resolveTabHref` falls back to the first page in the section. That fallback walks the navigation tree, so it only sees pages that come from content.

This means a tab pointing at a route generated outside the content tree lands somewhere unexpected. The case I ran into is the changelog: once a project has a `type: changelog` entry, Blume generates `/changelog`, but that page isn't in the tree — so a `{ label: "Changelog", path: "/changelog" }` tab resolves to the newest entry's own page rather than the index. The same applies to a custom page under `pages/`.

This PR adds an optional `href` to `navigation.tabs`. When set, it's the tab's link target; when omitted, resolution works exactly as before, so no existing site changes behaviour.

```ts
navigation: {
  tabs: [
    { label: "Changelog", path: "/changelog", href: "/changelog" },
  ],
}
```

`NavTab` already carried an internal `href` for the resolved value, so this exposes that field in the config rather than adding a new concept. Declared hrefs are localized and based like `path`, so they work under i18n and `deployment.base`.

I wasn't sure whether the current landing behaviour is deliberate — I noticed useblume.dev's own Changelog tab lands on the newest release, so it may well be.
This keeps that as the default either way, and only changes what happens when a tab asks for something specific.

## Related Issues

None for the change itself — happy to open one first if you'd rather discuss the API before the code.

Unrelated, but worth flagging so it isn't mistaken for fallout from this PR: I couldn't run the full suite locally because `configured-integrations.test.ts` times out on my machine. I filed that separately as #121. It fails the same way with this branch stashed, so it isn't caused by these changes.

## Checklist

- [x] I've reviewed my code
- [x] I've written tests

## Additional Notes

Tests: a new case in `navigation.test.ts` for a declared href winning over resolution, plus the existing basePath and i18n tab tests extended to cover one.

Docs: a short passage under Tabs in `content/navigation.mdx`, and a note in `advanced/changelog.mdx` where the generated index is described.

On my own site I worked around this by adding an `index.mdx` to the changelog folder, which puts `/changelog` in the tree — but that also stops the generated timeline, so it isn't a workaround if you want to keep the default page.

What I ran locally: `bun run check`, `bun run typecheck`, and `bun test --coverage` over the suite with `configured-integrations.test.ts` excluded (1975 pass, coverage thresholds met). I couldn't include that one test for the reason under Related Issues, so CI is the first place the full suite runs against this branch.

## Summary by CodeRabbit

* **New Features**
  * Navigation tabs can now specify a custom link destination via `href`, which is preserved even when the sidebar would otherwise fall back to another page.
  * `href` is localized per locale and respects the deployment base path.
  * Empty-string `href` values are rejected; omit `href` to use the section’s normal fallback behavior.
* **Bug Fixes**
  * Prevents misdirected tab links for routes that aren’t part of the authored content tree (for example, the generated `/changelog` page).
* **Documentation**
  * Updated navigation/tab docs and the changelog guidance to show configuring a tab with an explicit `href` (e.g., `/changelog`).


haydenbleasel:
Reviewed — this is a well-built PR, thank you. Verified against the surrounding code:

- All three layout components already render `withBase(tab.href ?? tab.path)`, so the declared href plugs into the existing render path (and picks up `deployment.base`) with no component changes.
- Active-tab matching and sidebar scoping in `nav-utils.ts` read only `tab.path`, so declaring an href can't affect section scoping.
- The localize-then-rebase order matches what `path` gets, `withBasePath` is idempotent, and external URLs pass through both transforms.
- Changeset, docs location, and tests all line up with how the repo does things.

I pushed a small commit to your branch (5fff235) with two touch-ups from the review, so you don't have to round-trip:

- The href pass-throughs in `graph.ts` and `buildNavigation` wrote an explicit `href: undefined` key onto tabs that never declared one; a conditional spread keeps their shape unchanged.
- Added assertions pinning that an external declared href passes through localization and basePath rebasing untouched.

Will merge once CI is green on that.

haydenbleasel:
Merged — thanks for the contribution! The write-up made this an easy review: the problem, the API choice, and what you tested were all laid out clearly. This will go out with the next release.

PR #124, opened 2026-07-26
Rotate a collapsible group's chevron from its own state, not an ancestor's

### LINKED ISSUE #123: Nested sidebar chevron incorrectly inherits its parent group's open state
**Describe the bug**

The chevron of a nested sidebar group incorrectly inherits the open state of its parent group.

When the parent group is expanded but the nested group is collapsed, the nested group's chevron still points down. Its child navigation items remain hidden, so the chevron direction does not match its actual state.

Open:
Image
Close:

Image

**Blume version**

I am using version 1.1.4.

**To Reproduce**

1. Create a sidebar containing nested collapsible groups.
2. Run the documentation site with `blume dev`.
3. Expand the parent group.
4. Keep the nested group collapsed.
5. Observe that the nested group's chevron points down even though its child items are hidden.

**Expected behavior**

The chevron should reflect the state of its own group:

- An expanded group should point down.
- A collapsed group should point right.

Expanding a parent group should not rotate the chevrons of collapsed nested groups.

**Environment (please complete the following information):**

- Command: `blume dev`
- Runtime: Node.js 24.14.1
- OS: macOS 15.7.7
- Browser: chrome 150

h-kono-it:
## Description

A collapsible sidebar group's chevron rotated on Tailwind's `group-open:` variant, which compiles to
a descendant selector:

```css
.group-open\:rotate-90:is(:where(.group):is([open],:popover-open,:open) *){rotate:90deg}
```

`NavTree.astro` gives every collapsible group its own `class="group"`, so a nested group's chevron is
also a descendant of each ancestor's `.group`. Expanding a parent therefore rotated the chevrons of
its collapsed children — the arrow pointed down at items it was still hiding.

This scopes the rotation to the group's own `details`:

```diff
-<span class="shrink-0 text-muted-foreground transition-transform group-open:rotate-90">
+<span class="shrink-0 text-muted-foreground transition-transform [details[open]>summary_&]:rotate-90">
```

which compiles to `details[open]>summary .…{rotate:90deg}`, so each chevron reflects only the state of
the `details` it belongs to. Nothing else about the markup or the animation changes.

## Related Issues

Closes #123

## Checklist

- [x] I've reviewed my code
- [x] I've written tests

## Screenshots

From a site with `navigation.sidebar.display: "group"` and category groups nested under a parent
group. In both shots the parent group is expanded, and so is its second child — the other children
are collapsed.

**Before** — every child chevron points down, including the collapsed ones:

before

**After** — only the expanded child points down; the collapsed siblings point right:

after

## Additional Notes

`ui-coverage.test.ts` already asserts NavTree's classes by reading the component source, so the
regression check lives there: it pins the scoped selector and asserts the old `group-open:rotate-90`
is gone.

`group-open:` is still used elsewhere (`TypeTable.astro` via a named `group/tt`, and the
`rotate-180` chevrons in the header dropdowns) — those don't nest inside another group of the same
name, so they're unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed nested sidebar chevrons rotating when ancestor navigation groups are expanded.
  * Updated chevron/toggle indicators across nested disclosures so they reflect only their own expanded/collapsed state (sidebar groups, Tree folders, Accordion items, and schema “Show properties” toggles).
* **Tests**
  * Added UI coverage to ensure nested disclosure chevron rotation is properly scoped to each disclosure.

### DIFF
--- a/.changeset/nested-disclosure-indicators.md
+++ b/.changeset/nested-disclosure-indicators.md
@@ -1,0 +1,5 @@
+---
+"blume": patch
+---
+
+Fix nested `<Tree>` folder chevrons, nested `<Accordion>` chevrons, and a nested object schema's "Show properties" toggle reflecting an ancestor's open state instead of their own. All three rotated or flipped on Tailwind's `group-open:` variant, which matches any open ancestor `.group` — the same leak as the nested sidebar chevron — so a collapsed disclosure inside an expanded one showed an open indicator. Each indicator is now scoped to its own `details`.
--- a/.changeset/nested-sidebar-chevron.md
+++ b/.changeset/nested-sidebar-chevron.md
@@ -1,0 +1,5 @@
+---
+"blume": patch
+---
+
+Fix a nested sidebar group's chevron pointing down while the group is collapsed. The chevron rotated on Tailwind's `group-open:` variant, which matches any descendant of an open `.group` — and since every collapsible group in the tree is a `.group`, expanding a parent rotated the chevrons of its collapsed children too, so the arrow disagreed with the items it was hiding. The rotation is now scoped to the group's own `details`, leaving each chevron to reflect only its own open state.
--- a/packages/blume/src/components/content/AccordionItem.astro
+++ b/packages/blume/src/components/content/AccordionItem.astro
@@ -22,7 +22,7 @@ const accordionId = id ?? slugify(title);
 ---
 
 <details
-  class="not-prose group"
+  class="not-prose"
   data-blume-accordion
   id={accordionId}
   open={defaultOpen}
@@ -52,7 +52,7 @@ const accordionId = id ?? slugify(title);
       </span>
     </span>
     <Icon
-      class="text-muted-foreground transition-transform group-open:rotate-180"
+      class="text-muted-foreground transition-transform [details[open]>summary_&]:rotate-180"
       name="chevron-down"
       size={16}
     />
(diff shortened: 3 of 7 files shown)


haydenbleasel:
Thanks for the fix and the clear writeup — the scoping is exactly right, and the before/after shots made it easy to verify.

While reviewing, I found the same leak in three other components the "used elsewhere" survey missed, all unnamed `.group` disclosures that nest in practice:

- `TreeFolder.astro` — folders nest via `<slot />`, so a collapsed subfolder's chevron pointed down inside an open parent.
- `SchemaProperty.astro` — nested object schemas render a `details` inside a `details`, so a collapsed nested property showed **"Hide properties"** while its properties were hidden.
- `AccordionItem.astro` — exposed when accordions nest in MDX (or sit inside any other open `.group`, since the class is global).

I pushed 6f2c04b3 to this branch extending your fix:

- Applied the same `[details[open]>summary_&]:` scoping to all three components.
- Dropped the now-unconsumed `group` class from each `details` (including NavTree's) so nothing invites a `group-open:` variant back — no script or stylesheet selects `.group`, so it was purely the Tailwind hook.
- Broadened your regression test to pin all four components, asserting `group-open:` is gone and the scoped variant is present in each.
- Added a second changeset for the content/OpenAPI components.

Full suite passes (1977 tests, lint, typecheck). Nothing about your original NavTree change was altered beyond removing the dead class and rewording the inline comment to match.

haydenbleasel:
Merged — thanks for the contribution! Great diagnosis and a really well-documented PR. The fix will go out in the next patch release.

PR #126, opened 2026-07-28
Match nav back button radius

tensei:
## Summary

- match the full-width navigation back row to the `0.65rem` radius used by other sidebar navigation buttons
- add regression coverage for the shared sidebar-row radius

## Visual verification

Both captures use a temporary `hover:bg-white/40` background so the radius difference is easy to see. The product hover color is unchanged.

### Before

The hovered back row uses a 4px radius while the other sidebar navigation buttons use 10.4px.

Before: hovered navigation back row with a 4px radius and white 40 percent screenshot background

### After

The hovered back row now uses the same 10.4px radius as the other sidebar navigation buttons.

After: hovered navigation back row with a 10.4px radius and white 40 percent screenshot background

## Root cause

The full-row click-target change used Tailwind's generic `rounded` utility instead of the sidebar's `rounded-[0.65rem]` navigation-row radius.

## Validation

- `bun test packages/blume/test/ui-coverage.test.ts`
- `bun run --filter blume typecheck`
- commit hooks: 1,980 tests passed, workspace typechecks passed, package build passed
- rendered page-mode verification in the docs app

## Summary by CodeRabbit

* **Style**
  * Updated the full-width navigation back button and related navigation rows to use a consistent `0.65rem` border radius.
* **Tests**
  * Added UI coverage assertions to ensure the navigation back button keeps the expected styling/behavior and to guard against conflicting rounded styling combinations.

### DIFF
--- a/.changeset/match-nav-back-radius.md
+++ b/.changeset/match-nav-back-radius.md
@@ -2,4 +2,4 @@
 "blume": patch
 ---
 
-Match the full-width navigation back row radius to the other sidebar navigation buttons.
+Match every hoverable sidebar navigation row — the back rows, routed panel header, and flat group header — to the shared 0.65rem navigation radius.
--- a/.changeset/match-nav-back-radius.md
+++ b/.changeset/match-nav-back-radius.md
@@ -1,0 +1,5 @@
+---
+"blume": patch
+---
+
+Match the full-width navigation back row radius to the other sidebar navigation buttons.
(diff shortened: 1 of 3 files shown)


haydenbleasel:
Pushed 92fe489 on top: review found the same 4px radius on the sibling rows — the routed panel header's back chevron and title link, and the flat group header link all still paired the generic `rounded` with `bg-muted`. They now use the shared `rounded-[0.65rem]`, and the new test asserts no NavTree class mixes bare `rounded` with a muted row background (it flags all three pre-fix offenders). Changeset wording broadened to match.

Issue #125, opened 2026-07-27
Default `orama` provider returns zero results for CJK content


h-kono-it:
**Describe the bug**

The default `orama` provider can't match Japanese text. Both the indexed content and the query tokenize to an empty array, so every Japanese query returns no hits — only terms that contain ASCII (e.g. "GDPR") are findable. Nothing errors; the dialog simply reports no results, so it reads as "the index is empty" rather than as a tokenizer mismatch. This should apply to Chinese and Korean too, though I have only measured Japanese.

`buildOramaIndex` builds the database from the schema alone (`packages/blume/src/search/orama-index.ts`):

```ts
const db = create({ schema: SCHEMA });
```

With no `components.tokenizer`, Orama defaults to `language: "english"`, whose splitter is `/[^A-Za-zàèéìòóù0-9_'-]+/gim`. Japanese characters fall outside that class, so they are all treated as delimiters and the string collapses to nothing. `i18n.defaultLocale` doesn't reach the tokenizer — it drives the UI strings and the `locale` enum filter — so declaring the site as Japanese doesn't change the outcome. Since `orama-index.ts` is shared, the MCP `search_docs` tool is affected in the same way as the client dialog.

**Blume version** 1.2.0

**To Reproduce**

1. Build a site with Japanese content on the default search provider. No particular configuration is needed — `create` is called with the schema alone, so nothing in the config can change the tokenizer.
2. Open the search dialog and type any Japanese word that appears in the content.
3. No results. The same word typed as part of an ASCII-bearing term (e.g. "GDPR") does return results.

My own site does set `i18n.defaultLocale: "ja"`, but that isn't what triggers this — I mention it only because declaring the site's language is the thing a user would expect to fix it.

Measured against that site's generated `blume-search.json` (65 pages, Japanese content), loading it into Orama with the same schema and boosts `orama-index.ts` uses:

| query | tokens produced | hits |
| --- | --- | --- |
| ポイント (points) | `[]` | 0 |
| 個人情報 (personal data) | `[]` | 0 |
| 退会 (account deletion) | `[]` | 0 |
| GDPR | `["gdpr"]` | 4 |

Page titles collapse the same way — `tokenize("法務に相談するときの準備リスト")` returns `[]` — so even an exact title match finds nothing.

**Expected behavior**

A Japanese term finds the pages containing it. If CJK support is out of scope for the bundled provider, then documenting the limitation would already help a lot, since the current failure mode is silent.

**Possible directions**

[...]

PR #132, opened 2026-08-01
Index CJK compound terms by character bigrams

h-kono-it:
## Description

Follow-up to #125. The segmenting tokenizer fixed the "zero results" failure, and I moved my site back to the default `orama` provider expecting to drop my Pagefind workaround. Comparing the two side by side, though, Japanese compound terms still ranked badly enough that I kept the workaround — so I looked into why.

Segmenting cuts a compound term into its parts: 資金決済法 becomes 資金 / 決済 / 法. Orama then scores a bag of words, so adjacency is gone, and a hub page mentioning many terms shallowly carries more of those fragments than the page any single term is about. On my site (65 Japanese pages, a reverse-lookup legal reference) **every** law-name query returned the site's index page first, and that page does not contain the term. Hit counts landed around 60 of 65 pages, because a fragment as common as 法 matches nearly everything.

The clearest case is 「出会い系」 (a category of dating service, and a named law). It segments to 出会い / 系, and my category names all end in 系 (お金・決済系, 通信・プライバシー系, …):

```
「出会い系」   before: 16 results, led by accounting and privacy pages
              after:   2 results — exactly the 2 pages that contain the term
```

This PR indexes runs of Han, Hiragana and Katakana as overlapping character bigrams, dropping the whole-segment tokens so a lone fragment can't match on its own. That's what Lucene's CJK analyzer does, and it also absorbs segmentation drift — under Pagefind's segmentation 「景品表示法」 reached only 1 of the 7 pages containing it, for the same underlying reason.

Because every 2-character window is individually common, ranking alone isn't enough: queries on a bigrammed index run a strict pass (`threshold: 0`) first, and fall back to the any-token default when no page carries the term in full, so sentence-like queries still return their closest pages instead of nothing. `queryOramaIndex` decides this from `db.tokenizer.language`, so no signature and no caller changed.

Hangul and Thai are deliberately untouched — Korean is written with spaces, and Thai has no comparable bigram convention — so `ko`/`th` sites keep exactly today's behaviour.

### Measurements

My site's real index. 20 reverse-lookup queries (law names, okurigana compounds, katakana, Latin acronyms) plus 6 sentence-like ones. "top-3 correct" counts how many of the first three results actually contain the queried term; MRR is over the rank of the first correct hit. Same corpus, same queries, same scoring for every row.

| | top-3 correct | MRR | sentence queries returning nothing |
|---|---|---|---|
| before (1.2.1) | 36/53 (68%) | 0.762 | 0/6 |
| bigrams only | 43/53 (81%) | 0.975 | 0/6 |
| **this PR** | **53/53 (100%)** | **1.000** | **0/6** |

After the change, hit counts equal the ground-truth count exactly for all 20 queries — 資金決済法 returns 9 pages and 9 pages contain it, 下請法 4 and 4, 出会い系サイト規制法 2 and 2.

## Related Issues

Follow-up to #125, which is already closed — there's no open issue for this, so nothing to auto-close.

## Checklist

- [x] I've reviewed my code
- [x] I've written tests

Not a UI change — the before/after is in the result lists and the table above.

## Additional Notes

**The two halves are separable, if you'd rather take only one.** The tokenizer alone carries most of the ranking win (MRR 0.762 → 0.975); the strict pass is what closes the remaining gap. If it feels like too much policy for one PR, I'm happy to split it out and leave the query side as it is. Note that sentence-like queries return nothing under a plain `threshold: 0`, which is why the fallback exists rather than just passing the threshold through.

**Calibration against Pagefind.** I ran `pagefind_extended`, which segments CJK natively, over the same corpus and query set: **43/53 (81%), MRR 0.956** — the same top-3 rate as bigrams alone. So this puts the default provider ahead of the workaround I had been using.

**Verification.** 6 new cases in `mcp.test.ts` alongside the existing CJK ones: compound ranking against a hub page, the okurigana case, the lone-character path, the fallback, and Latin/Hangul/Thai staying whole. `bun test --coverage` keeps `orama-index.ts` at 100%, and the existing zh/ko/th and mixed-locale assertions pass unchanged. The numbers above come from running this branch's `orama-index.ts` directly against my site's real `blume-search.json`, not from the fixtures.

**What I have not verified.** I only have a Japanese corpus. Chinese takes the same Han path and should behave the same way, but I have not measured it. Korean and Thai are unchanged by design — if you'd rather they were bigrammed too, that's a call I didn't feel able to make.

**One judgement I'd like your read on.** Katakana currently joins the kanji/kana run. Keeping it whole instead scores identically on my corpus and differs only at the edges: as a whole word it returns a clean zero for a term that isn't in the docs, while bigrammed it tolerates orthographic variation (a query for クーリングオフ finds a page written クーリング・オフ). I went with bigrams for the variant tolerance, since katakana loanwords vary a lot in Japanese technical writing, but I can see the argument the other way.

## Summary by CodeRabbit

* **New Features**
  * Improved Japanese and Chinese search relevance through overlapping character-pair matching and ranking.
  * Added fallback matching when full compound-term matches are unavailable.
  * Preserved existing search behavior for Korean, Thai, Latin, and other text.

* **Documentation**
  * Updated search documentation to explain language-specific matching behavior.

* **Tests**
  * Added coverage for Japanese compounds, okurigana, supplementary characters, single-character queries, fallback matching, and multilingual terms.


h-kono-it:
Both review points were real — thanks. Pushed a follow-up commit.

**Keeping indexing and querying aligned.** This one was my mistake, and the description above was wrong because of it: `tokenize` bigrammed Han and kana runs on *every* segmented index while the strict pass ran only for `ja`/`zh`, so Han or kana on a `ko`/`th` index was indexed as bigrams and then matched loosely — precisely the combination the strict pass exists to prevent. I took the first of the two suggested options and gated bigram generation, so both paths now read `BIGRAM_LANGUAGES` and can't drift apart. `ko`/`th` really are untouched now, which is what the description claimed all along. Added a regression test with Han content in a `ko` index, asserting a fragment of the compound still matches it as it does today.

**Supplementary-plane ideographs.** Also right, and there was a second half to it. Widening the class alone would have made things worse, because `addBigrams` cut its windows by code unit:

```
𠮟責  (U+20B9F U+8CAC)
  current class      → no match, falls through to a whole-segment token
  widened class only → ["𠮟", "\udf9f責"]   ← the surrogate pair is split
```

So the class now uses property escapes *and* the windows are cut by code point. One caveat worth recording: `ー` (U+30FC) is not `Script=Katakana`, and the iteration marks belong to no script of their own, so switching to bare property escapes would have quietly dropped long vowels from katakana words — `々〆〇ーﾞﾟ` are kept as literals for that reason. Added a test on 𠮟責, which isn't a curiosity: it's the 常用漢字表 form of しかる.

Verification is unchanged from the first commit: 1998 tests pass, `orama-index.ts` stays at 100% coverage, and re-running the branch against my site's real index still gives 53/53 top-3 correct and MRR 1.000, so neither change cost anything on the measurements above.

haydenbleasel:
Thanks @h-kono-it — this is an excellent PR, and the measurements made the case for taking both halves together. Keying the strict pass off `db.tokenizer.language` is a nice touch: on the no-`Intl.Segmenter` fallback path the language is not a bigram one, so the strict pass is correctly skipped and index and query policy can't drift apart.

I verified on the branch: all tests pass, `orama-index.ts` stays at 100% coverage, and the Unicode claims in the comments hold up empirically — ﾞﾟー are `Script=Common` so the extra literals are needed, while halfwidth katakana and the kana iteration marks are already covered by the property escapes.

On the katakana question: I'd keep it bigrammed, as you have it. It matches Lucene's CJKAnalyzer, and the variant tolerance (クーリングオフ ↔ クーリング・オフ) is worth more in Japanese technical writing than a clean zero for absent terms — which the strict pass softens anyway.

Rather than a review round trip I pushed a small follow-up commit (1e8930b2): the lone-character test now also asserts the out-of-reach case (a run-final 法 in 景品表示法。 is unreachable by the query 法 — the same tradeoff Lucene makes), `addBigrams` documents that tradeoff, the changeset mentions that multi-word Latin queries on a bigrammed index now prefer pages carrying every word, and a test name switches to American spelling per repo convention.

Merging — thanks again for the thorough writeup.

PR #160, opened 2026-08-04
feat(search): allow asset and inline SVG icons on popular links

### LINKED ISSUE #159: search.popular icons: support image paths and inline SVG like nav
## Problem

`search.popular` only accepts built-in Lucide names. An image path or inline SVG falls through to the file glyph, and diagnostics warn about it.

Nav icons already accept all three shapes (built-in name / path / inline SVG). Popular links are resolved on the server before the Cmd+K island runs — so there's no real reason they can't take the same inputs. The island already injects prebuilt markup; `iconSvg()` just never handled assets.

Hit this on a hub that wants real product marks in the Cmd+K empty state instead of Lucide stand-ins.

## Expected

Same icon shapes as nav:

```ts
search: {
  popular: [
    { href: "/codemap", label: "Codemap", icon: "/icons/codemap.svg" },
    { href: "/persist", label: "Persist", icon: "<svg …></svg>" },
    { href: "/about", label: "About", icon: "info" },
  ],
}
```

## Notes

- Docs currently call out the limitation under search config.
- Small fix: resolve assets to `` / pass through inline SVG on the server, drop the asset warning.

SutuSebastian:
## Summary
- Resolve `search.popular` image paths and inline SVG to markup on the server (same shapes as nav `Icon`), so Cmd+K can show real marks instead of only Lucide names.
- Drop the diagnostic that treated assets as unsupported; update search docs + types.

Fixes #159

## Test plan
- [x] `bun test packages/blume/test/popular-icon.test.ts packages/blume/test/nav-diagnostics.test.ts`
- [x] pre-commit coverage / build
- [ ] Config `search.popular` with `/logo.svg` and an inline `<svg>` — icons render in Cmd+K empty state
- [ ] Built-in name still works; unknown name still warns

## Summary by CodeRabbit

* **New Features**
  * Popular search links now support built-in icons, image paths or URLs, and inline SVG icons.
  * Image icons automatically respect the site’s base path.
  * Unrecognized or missing icons fall back to the default file glyph.

* **Bug Fixes**
  * Valid image and SVG icons are no longer flagged as invalid.
  * Image sources are safely escaped when rendered.

* **Documentation**
  * Updated search configuration guidance to describe the expanded icon support.

### DIFF
--- a/.changeset/search-popular-asset-icons.md
+++ b/.changeset/search-popular-asset-icons.md
@@ -1,0 +1,5 @@
+---
+"blume": patch
+---
+
+Allow `search.popular` icons to use image paths/URLs and inline SVG, matching nav icons. Markup is resolved on the server before the Cmd+K island runs.
--- a/apps/docs/content/docs/configuration/search.mdx
+++ b/apps/docs/content/docs/configuration/search.mdx
@@ -27,9 +27,7 @@ search: {
 },
 ```
 
-Each entry takes an `href` (internal route or external URL) and a `label`, plus an optional `icon` — a built-in icon name shown beside the label, defaulting to a file glyph. Omit `popular` or leave it empty to keep the sidebar fallback.
-
-Unlike icons elsewhere in Blume, `icon` here must be a built-in name: these rows render in a client island, so an image path or inline SVG isn't supported and falls back to the file glyph.
+Each entry takes an `href` (internal route or external URL) and a `label`, plus an optional `icon` — a [built-in icon](/docs/content/components#icon) name, image path/URL, or inline SVG (same as nav icons), defaulting to a file glyph. Omit `popular` or leave it empty to keep the sidebar fallback.
 
 Write `href` as if the site were mounted at the root — a `basePath` is applied for you, the same as `navigation.featured`. External URLs pass through untouched.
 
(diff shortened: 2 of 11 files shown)


SutuSebastian:
Addressed the CodeRabbit note on invalid markup vs the resolver:

- Shared `isInlineSvg` / `isAssetIcon` already reject values like `` (warn + file-glyph fallback).
- Added tests for that case in `nav-diagnostics` and `popular-icon`.

haydenbleasel:
Pushed 700ac87 with two small follow-ups from review:

- **Inline SVG popular icons are now wrapped in a fixed-size span** (`display:inline-flex;width:16px;height:16px`, the same wrapper `Icon.astro` uses). Without it, a viewBox-only `<svg>` renders at the browser-default 300×150 inside the Cmd+K rows — the `` and built-in branches already carried explicit sizes.
- **Shape-aware diagnostic for markup icons**: a string like `` (or a truncated `<svg`) now warns that it isn't a complete inline `<svg>` element, rather than the misleading "isn't in Blume's icon set". Applies to both nav and popular-link icons via the shared helper.

Tests updated to match; full coverage suite is green. Thanks for the PR!

SutuSebastian:
Awesome! Good extra layers

PR #161, opened 2026-08-04
fix(deploy): serve a Cloudflare server build's artifacts from dist/client

### LINKED ISSUE #155: deployStaticDir() returns dist/ on Cloudflare server builds: every generated artifact 404s, and blume audit reads client/ as a route segment
On a Cloudflare **server** build, every artifact Blume writes through `deployStaticDir()` lands one directory above the only directory the Worker serves. `robots.txt`, `sitemap.xml`, `llms.txt`, `llms-full.txt`, `agent-readability.json` and `.well-known/api-catalog` all 404 in production, and `blume audit` reports 118 errors that are all false.

I think it's one clause in `deployStaticDir`.

## Repro

```ts
// blume.config.ts
export default defineConfig({
  deployment: {
    site: 'https://example.com',
    adapter: 'cloudflare',
    output: 'server',
  },
});
```

```
blume build
```

## Root cause

`src/deploy/adapter-output.ts`:

```ts
export const deployStaticDir = (config, context) => {
  const { adapter, output } = config.deployment;
  if (output === "server" && adapter === "vercel") {
    return join(context.root, ".vercel", "output", "static");
  }
  const dist = context.distDir ?? join(context.root, "dist");
  if (output === "server" && adapter === "node") {
    return join(dist, "client");
  }
  return dist;                                    // ← cloudflare + server lands here
};
```

The doc comment above it explains the reasoning:

> Netlify publishes `dist/` itself and Cloudflare serves the `outDir` root, so every other build serves `dist/`.

That's true for a Cloudflare **static** build. It isn't for a Cloudflare **server** build: `@astrojs/cloudflare` writes its own `dist/server/wrangler.json` pointing the asset binding at the *client* directory —

```json
{ "binding": "ASSETS", "directory": "../client" }
```

— so `dist/client` is served and `dist/` is not. Cloudflare + `server` needs the same treatment as `node` + `server`, for the same reason.

## Symptom 1 — the artifacts 404

Everything the build summary reports as generated is unreachable:

```
$ blume build
✔ Generated llms.txt and llms-full.txt
✔ Generated sitemap.xml
✔ Generated agent-readability.json
✔ Generated .well-known/api-catalog (RFC 9727)
✔ Built to /path/to/dist

$ find dist -maxdepth 2 -type f -not -path 'dist/client/*' -not -path 'dist/server/*'
dist/.well-known/api-catalog
dist/agent-readability.json
dist/llms-full.txt
dist/llms.txt
dist/sitemap.xml

$ find dist/client -maxdepth 2 \( -name 'llms*.txt' -o -name 'agent-readability.json' -o -name 'api-catalog' \)
(nothing)
```

Every line of that summary is true, and none of those files is served. (`robots.txt` would be in the same list — it's absent above only because this project sets `seo.robots: false` and ships its own.)

## Symptom 2 — `blume audit` is unusable on this adapter

`src/audit/run.ts` resolves its root through the same function, so it crawls `dist/` and reads `client/` as a **route segment**. Every page is audited as `/client/docs/…`:

```
blume audit  12 pages · dist · offline
912 audits · 118 errors · 54 warnings · 1 note

✖ Referenced script, style, or asset is missing from the build   55 pages
    /client/404
✖ Canonical points to a broken or redirecting page               11 pages
    /client/docs/api
```

Both are false. The asset it calls missing is present and serves 200:

```
/_astro/Fonts.CMeh-ou_.css   PRESENT in dist/client   → 200 from the Worker
```

and every canonical it calls broken resolves 200. Because the crawl root is wrong, relative asset references resolve against `dist/` (where `_astro/` doesn't exist) and page paths gain a `/client` prefix that no canonical will ever match. There's no flag to point the audit elsewhere, so on this adapter the command can't be used at all — which also means it can't be wired into CI.

## Symptom 3 — the chain

`agent-readability.json` is itself one of the unreachable files, and it advertises another one:

```json
{ "artifacts": { "apiCatalog": "https://example.com/.well-known/api-catalog" } }
```

So the manifest an agent reads to discover everything else 404s, and the URL it would have pointed at 404s too.

## Suggested fix

```diff
-  if (output === "server" && adapter === "node") {
+  if (output === "server" && (adapter === "node" || adapter === "cloudflare")) {
     return join(dist, "client");
   }
```

Happy to open a PR if you'd like it in that shape. Worth a look at whether `surfaceAdapterOutput`'s `ADAPTER_OUTPUT_PATHS` wants a Cloudflare entry for the same reason, though I haven't hit anything there.

## Workaround

A post-build script that moves the four files with no Astro-native substitute (`llms.txt`, `llms-full.txt`, `agent-readability.json`, `.well-known/api-catalog`) from `dist/` into `dist/client`. `robots.txt` and `sitemap.xml` don't need it — `seo.robots: false` plus a `public/robots.txt`, and `@astrojs/sitemap`, both write into Astro's own client output. There's no seam for the other four: `buildLlmsFiles` needs a build-time `BlumeProject` an injected route can't reach, and Blume writes them after `astro build` returns, so no integration hook fires late enough.

## Environment

| | |
|---|---|
| blume | 1.3.1 (also 1.2.1) |
| astro | 7.1.6 |
| @astrojs/cloudflare | 14.1.7 |
| wrangler | 4.118.0 |
| runtime | bun 1.3.14, macOS |

[Live site on the workaround](https://sigiro.com/llms.txt)

akshatagarwl:
Fixes #155

## Root cause

`@astrojs/cloudflare@14` declares `preserveBuildClientDir: true`, so it keeps Astro's `dist/client` + `dist/server` split and rewrites the ASSETS binding in the `dist/server/wrangler.json` it generates to point at `../client`. The deployed Worker therefore serves `dist/client` and nothing above it.

`deployStaticDir()` still returned `dist/` for cloudflare + server (the older Pages-style layout), so every artifact Blume writes — `robots.txt`, `sitemap.xml`, `llms.txt`, `llms-full.txt`, `agent-readability.json`, `.well-known/api-catalog` — landed one directory above the served root and 404'd in production. `blume audit` crawled the same wrong root and read `client/` as a route segment, producing ~118 false errors.

## Changes

1. `packages/blume/src/deploy/adapter-output.ts` — `deployStaticDir` returns `join(dist, "client")` for cloudflare as well as node on a server build.
2. `packages/blume/src/cli/commands/build.ts` — `isolatedStaticDir` mirrors the same clause; without it an isolated verify build points `--analyze` / the budget gate at the wrong root.
3. Doc comments on both functions — the old wording ("Cloudflare serves the `outDir` root") described the pre-v14 adapter layout. Both now explain the `preserveBuildClientDir: true` + `dist/server/wrangler.json` ASSETS binding reason a server build differs, and that the `outDir` root remains correct for a Cloudflare *static* build.

Static Cloudflare builds are unchanged — the existing `output === "server"` guard already leaves them on `dist/`.

## Fixed transparently, no separate edits

The Pagefind search index (`publishBuildArtifacts`) and the `blume audit` crawl root resolve `staticDir` through the two functions above, so they are corrected by the same change. `_redirects` and `_headers` turn out to be unaffected either way — both emitters early-return unless `output === "static"`, so a Cloudflare server build never wrote them in the first place. *(Corrected by maintainer — an earlier revision of this description listed them as fixed; the changeset was updated to match.)*

`ADAPTER_OUTPUT_PATHS` is deliberately unchanged — cloudflare emits into `dist/` at the project root, so it needs no surfacing entry. This is only about *which* directory inside `dist/` is served.

## Test plan

- Extended `packages/blume/test/adapter-output.test.ts`: cloudflare + server resolves to `/proj/dist/client`; cloudflare + static still resolves to `/proj/dist`.
- Extended `packages/blume/test/build-budget-gate.test.ts`: same two cases through `isolatedStaticDir`.
- `bun test` in `packages/blume` — 2153 pass, 0 fail across 113 files.
- `bun run typecheck` — clean (the two `BLUME_NAV_INDEX_TITLE_MISMATCH` warnings in `apps/docs` content are pre-existing and unrelated).
- `bun run check` (ultracite) — 0 warnings, 0 errors, formatting clean.
- Verified the adapter behaviour against the published `@astrojs/cloudflare@14.1.7` tarball rather than inferring it: `preserveBuildClientDir: true` at `dist/index.js:341`, and the `astro:build:done` hook rewriting `wrangler.assets.directory` to `relative(build.server, originalClientDir)` at `dist/index.js:407-418`.

Not verified: no live `wrangler deploy` was run — the fix is asserted against the adapter's emitted config rather than a deployed Worker.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Cloudflare server deployments to correctly serve static assets from the client output directory.
  - Preserved the existing output location for Cloudflare static builds.
  - Updated deployment handling to align static asset paths across supported server and static build modes.

- **Tests**
  - Added coverage verifying static asset paths across Cloudflare server and static deployment modes.


haydenbleasel:
Thanks for the fix and the unusually thorough write-up — verifying against the published adapter tarball made this easy to confirm. I reproduced the `preserveBuildClientDir: true` behavior against `@astrojs/cloudflare@14.1.7` independently and everything checked out, so I've pushed a small follow-up commit (dc4e61e8) directly to your branch rather than round-tripping review comments:

- Extracted the duplicated `node`/`cloudflare` server-build clause from `deployStaticDir` and `isolatedStaticDir` into a shared `servesClientSubdir` predicate, so the client/server-split explanation lives in one doc comment instead of two that could drift.
- Removed `_redirects` and `_headers` from the changeset's list of artifacts that 404'd: both emitters early-return unless `output === "static"`, so a Cloudflare server build never wrote them in the first place — they were unaffected by the wrong served root. (I made the matching correction to the PR description, since the changeset text becomes the CHANGELOG.)

Full suite (2153 tests), typecheck, and formatting all pass on the updated branch, and CI is green. Will merge shortly.

PR #165, opened 2026-08-05
fix(deploy): emit _headers on a Cloudflare server build

akshatagarwl:
## Problem

A Cloudflare **server** build serves no homepage `Link` header, and no `Content-Type` on the extensionless well-known files. Blume generates the whole agent-discovery surface for these sites and then never advertises it — an API catalog goes out with no media type instead of `application/linkset+json`.

Reproduced on a live site (`output: 'server'`, `adapter: 'cloudflare'`):

```
$ curl -sD- -o /dev/null https://sigiro.com/ | grep -i '^link:'
(nothing)

$ curl -sD- -o /dev/null https://sigiro.com/.well-known/api-catalog | grep -i content-type
(nothing)
```

`buildHomeLinkHeader()` builds the correct value. It has exactly three consumers, and each excludes this deployment:

| Consumer | Guard | Cloudflare server |
| --- | --- | --- |
| `cli/commands/build.ts` → `_headers` | returns early unless `output === "static"` | ✗ |
| `cli/commands/build.ts` → `.vercel/output/config.json` | Vercel adapter only | ✗ |
| `astro/integration.ts` → `res.setHeader("Link", …)` | mounted on `astro:server:setup` | ✗ dev only |

So Netlify/Cloudflare **static**, Vercel, and `blume dev` all get the header; Cloudflare server and Node server get nothing.

The root cause is a conflation, and the existing comment shows it:

> Server adapters set the Content-Type on the Response directly, so this is static-only.

That is true of the **charset** rules in `_headers` — the runtime endpoint sets those itself. It is not true of the `Link` rule or the well-known media types, which no server adapter sets. One correct conclusion about part of the file was applied to all of it.

## Fix

Gate on a new `readsHeaderFiles()` instead of `output === "static"`: true for any static build, plus a Cloudflare server build, whose Worker serves `dist/client` through its ASSETS binding and honours `_headers` from that directory exactly as Pages does.

- **Node server stays excluded** — the standalone server's static handler ignores `_headers`, so writing it would be inert, and reporting a header that is not applied is worse than omitting it.
- **Vercel stays excluded** — its headers arrive through the routing config, which this would duplicate.

Two further corrections came out of testing, and **without them the fix does not fire**:

1. **The user opt-out is now read from `public/_headers`, not `dist`.** `@astrojs/cloudflare` writes its own `_headers` during the build (an immutable `Cache-Control` for `/_astro/*`), so the `existsSync(dist/_headers)` check was reading an adapter-generated file as a user opt-out and skipping silently. The doc comment already described the intent as "a `_headers` shipped in `public/`" — the check just tested the wrong location.
2. **An existing `_headers` in the output is preserved and appended to**, so the adapter's caching rule and Blume's discovery rules coexist.

## Verification

Unit tests cover the predicate across all four adapters in both output modes. But the interesting part is the end-to-end run, because the unit tests would have passed while the feature stayed broken — correction 1 above was only visible against a real build.

Built a real Cloudflare server site against this branch. Before, `dist/client/_headers` held only the adapter's rule. After:

```
/_astro/*
  Cache-Control: public, max-age=31536000, immutable
/docs/*.md
  Content-Type: text/markdown; charset=utf-8
/docs/*.mdx
  Content-Type: text/markdown; charset=utf-8
/*.txt
  Content-Type: text/plain; charset=utf-8
/
  Link: </.well-known/api-catalog>; rel="api-catalog"; type="application/linkset+json", </agent-readability.json>; rel="describedby"; type="application/json", </llms.txt>; rel="describedby"; type="text/plain", </index.md>; rel="alternate"; type="text/markdown"
/.well-known/api-catalog
  Content-Type: application/linkset+json
```

The adapter's `Cache-Control` rule survives at the top, the `Link` header carries all four relations, and the API catalog gets its registered media type.

`bun run check` clean, `bun run typecheck` 3/3, `bun test` **2164 pass / 0 fail**.

## Note

The charset rules remain redundant on a server build, since the runtime endpoint sets `Content-Type` on the Response. They are harmless — a static-asset rule only applies to a file actually served from that directory — and splitting the file into "charset half" and "discovery half" seemed worse than emitting a few inert lines. Happy to gate them if you would rather.

Same shape as #161: the Cloudflare server path being an afterthought next to static and Vercel.

## Summary by CodeRabbit

* **New Features**
  * Cloudflare server deployments now generate `_headers` files with AI discovery and media-type rules.
  * Existing deployment-generated header rules are preserved and augmented automatically.
  * Add a `public/_headers` file to opt out of automatic header generation.

* **Bug Fixes**
  * Header generation now applies consistently to supported static and Cloudflare builds while remaining disabled for Node server and Vercel deployments.


haydenbleasel:
Great find and a careful writeup — thank you. I verified the load-bearing claims: `emitHeaderFiles` receives `deployStaticDir(...)`, which resolves to `dist/client` on a Cloudflare server build (where the ASSETS binding reads `_headers`), and the generated Astro config points `publicDir` at the project's `public/`, so the relocated opt-out check reads the right place. On the note at the end: agreed, leave the charset rules in — inert rules in one file beat splitting it.

I pushed a2d4643e onto the branch with three review follow-ups:

- **Tests for the two corrections the fix depends on.** `readsHeaderFiles` was covered, but the `public/_headers` opt-out location and the append-to-adapter-file behavior were only verified against a real build — exactly the parts that, as the PR description says, silently kept the old code from firing. `emitHeaderFiles` is now exported and exercised in subprocesses (`test/build-header-files.test.ts`), following the pattern of the other command suites.
- **Comment precision in `readsHeaderFiles`.** Only a Vercel *server* build is excluded; a static build on the Vercel adapter still returns true and gets the (ignored, inert) file, matching the old `output === "static"` gate. The comment now says so.
- **Changeset note for static builds.** The append behavior is a semantic change there too: a `_headers` that reached `dist` some way other than `public/` previously suppressed generation and is now appended to.

Also normalized honour → honor in the new comments (repo uses American spelling). `bun run check` clean, typecheck clean, 2168 tests pass. Will merge once CI is green.

Issue #166, opened 2026-08-05
Cloudflare server builds ignore Accept: text/markdown — middleware alone is not enough, ASSETS pre-empts it


akshatagarwl:
**Is your feature request related to a problem? Please describe.**

A deployed Cloudflare **server** build ignores `Accept: text/markdown`. The `.md` variants Blume generates are reachable only at their explicit URLs, so an agent that sends the header gets HTML.

On a live site (`output: 'server'`, `adapter: 'cloudflare'`):

```console
$ curl -sD- -o /dev/null -H 'Accept: text/markdown' https://sigiro.com/docs/quickstart/ | grep -i content-type
content-type: text/html

$ curl -sD- -o /dev/null https://sigiro.com/docs/quickstart.md | grep -i content-type
content-type: text/markdown; charset=utf-8
```

The content is there and correct — only the negotiation is missing. Same underlying shape as #155 and the `_headers` gap in #165: negotiation has two implementations and neither covers this deployment.

| Implementation | Guard | Cloudflare server |
| --- | --- | --- |
| `deploy/vercel-negotiation.ts` (`injectNegotiationRoutes`) | requires `.vercel/output/config.json` | ✗ Vercel only |
| `astro/integration.ts` (`negotiateMarkdown`) | mounted on `astro:server:setup` | ✗ dev only |

The comment on `negotiateMarkdown` states the assumption that does not hold here:

> Runs only under `blume dev` — in production the content pages are prerendered and served from the platform's static layer, which this middleware never fronts.

That is true, and it is the actual blocker — see below. It just leaves Cloudflare server builds with no negotiation at all, while `blume dev` and Vercel both have it.

**Describe the solution you'd like**

Astro middleware alone is **not sufficient**, and I would rather report that than open a PR that looks right and does nothing.

I implemented it and tested against a real Cloudflare server build. The middleware generated correctly (all 11 content routes, correct `deployment.base` handling), bundled into the Worker as `virtual_astro_middleware.mjs` (3.29 KiB), and `bun run typecheck` passed. The shared helpers behave correctly in isolation:

```
prefersMarkdown('text/markdown')            -> true
markdownVariantUrl('/docs/quickstart/', …)  -> /docs/quickstart.md
markdownVariantUrl('/', …)                  -> /index.md
```

[...]

PR #170, opened 2026-08-07
fix(deploy): keep configured redirects off the Cloudflare worker-first routes

### LINKED ISSUE #169: Cloudflare server builds serve configured redirects as permanent 301s: run_worker_first swallows them and _redirects is never consulted
On a Cloudflare **server** build, every configured redirect that happens to live under a content prefix is served as a permanent **301**, whatever status the config asked for. `blume.config.ts` says `status: 302`, the generated `dist/client/_redirects` says `302`, and the edge sends `301` — which browsers cache more or less indefinitely, so it cannot be walked back once shipped.

The two outputs Blume generates contradict each other.

### Mechanism

1. `emitRedirectFiles` skips a non-static build, so Blume's own `_redirects` is not involved. Instead `astro/templates.ts` writes `redirects` into the generated Astro config as `{ destination, status }` — correct, and the status is right there.
2. `@astrojs/cloudflare` sets `build.redirects: false` (no meta-refresh HTML) and instead appends every `type: "redirect"` route to `dist/client/_redirects` via `createRedirectsFromAstroRoutes`, preserving the exact status. On my site that file has 29 rules, all `302`, zero `301`. **This file is the only thing that serves a redirect with its configured status,** and only Cloudflare's static layer reads it.
3. `injectWorkerNegotiation` computes `assets.run_worker_first` from `buildRunWorkerFirstRules(options.routePaths, options.base)`, which groups by the **first path segment** of the content routes. One page at `/docs/reference` is therefore enough to emit `/docs` + `/docs/*`.
4. `/docs/*` swallows redirect paths like `/docs/api` and `/docs/api/run-query`. Inside a worker-first route Cloudflare hands the request to the Worker and **never consults `_redirects`**.
5. Astro answers instead. Its SSR redirect handler honors the configured status only when the destination resolves to a discrete route:

```js
// astro/dist/core/redirects/render.js
function computeRedirectStatus(method, redirect, redirectRoute) {
  return redirectRoute && typeof redirect === "object"
    ? redirect.status
    : method === "GET" ? 301 : 308;
}
```

`redirectRoute` is `routeMap.get(destination)`, and `routeMap` is keyed by *route patterns*. Blume serves every content page from `[...slug].astro`, so a destination like `/docs/reference` never resolves to a discrete route and `redirectRoute` is always `undefined`. Confirmed in the shipped bundle — the manifest entry carries no `redirectRoute` key at all:

```json
{
  "type": "redirect",
  "route": "/docs/api/get-trace",
  "redirect": { "status": 302, "destination": "/docs/reference" },
  "prerender": false
}
```

So the fallback fires and a GET gets **301**. This is not an occasional edge case: under Blume it is the guaranteed outcome for every redirect that reaches the Worker.

`buildRunWorkerFirstRules` only receives `routePaths`, so it structurally cannot know a redirect lives under a content prefix. At the call site in `cli/commands/build.ts`, though, `config` is already in scope — `config.redirects` is available and simply unused. The mechanism for the fix also already exists: Blume emits negative rules today (`!/docs/*.md`, `!/docs/*.mdx`) to keep the raw Markdown URLs on the static layer. Redirects need exactly the same treatment, for a stronger reason — `_headers` is cosmetic on a worker-first route, `_redirects` is load-bearing.

### Reproduction

Content at `/docs/reference` (so `basePath: "/docs"`), plus:

```ts
redirects: [{ from: '/api', to: '/reference', status: 302 }]
```

`dist/client/_redirects` gets `/docs/api /docs/reference 302`, and `run_worker_first` gets `/docs/*`. Then, against the generated bundle with `wrangler dev --config dist/server/wrangler.json`:

| `run_worker_first` | `/docs/api` |
| --- | --- |
| `["/","/docs","/docs/*","!/docs/*.md","!/docs/*.mdx"]` (what Blume emits) | **301** |
| same + `["!/docs/api","!/docs/api/*"]` | **302** |

The only difference is whether the redirect is exempted, which isolates it from any question about the redirect itself. I verified the same discrimination on a live deployment before working around it: every path inside `run_worker_first` returned 301 (`/docs/api`, `/docs/api/investigate-services`, `/docs/api/run-query`, `/docs/api/v1/query/post`) while every path outside it returned 302 — including `/sitemap.xml`, which is same-origin and so rules out an internal-vs-external explanation, plus `/openapi.json` and `/twitter`.

### Workaround

A manual negative rule in the project's `wrangler.jsonc`, since `mergeRunWorkerFirstRules` merges user rules ahead of the generated ones:

```jsonc
"assets": {
  "run_worker_first": ["!/docs/api/*", "!/docs/api", "/", "/docs", "/docs/*"]
}
```

The positives have to be restated because Wrangler rejects an all-negative list (`Only negative run_worker_first rules were provided; must provide at least 1 non-negative rule`). This should not be the user's job — the redirects are configured in `blume.config.ts` and Blume already knows them at the point where it writes the rules.

### Suggested fix

Pass the configured redirects (based the way the host platform matches them, i.e. `applyBaseToPlatformRedirects`) into `buildRunWorkerFirstRules`, and emit a negative rule for each one a positive rule would otherwise claim. Points worth handling deliberately:

- Wrangler rejects an all-negative list, so at least one positive must survive.
- Wrangler rejects a rule another same-polarity glob makes redundant (`validateStaticRoutingRules`), so `!{path}` must pair with `!{path}/*` rather than `!{path}/`.
- There is a 100-rule / 100-char ceiling. A retired section can collapse to a `{path}` + `{path}/*` pair, but only when no content route lives underneath — otherwise the glob would take that route off the Worker and break its Markdown negotiation.
- A path that is also a content route must never be exempted.
- The exemptions matter *more* in the `FALLBACK_RULES` path, not less, since `/*` claims every redirect.

Happy to send a PR — I have one working with tests.

### Environment

blume `1.3.1` (main @ `0895508`), astro `7.1.6`, `@astrojs/cloudflare` `14.1.7`, wrangler `4.118.0`, `deployment.output: "server"`, `basePath: "/docs"`.

akshatagarwl:
Fixes #169.

A configured redirect that happens to live under a content prefix shipped as a permanent **301** on a Cloudflare server build, whatever status the config asked for. `blume.config.ts` said `302`, the generated `dist/client/_redirects` said `302`, and the edge sent `301` — which browsers cache more or less indefinitely, so it cannot be walked back once shipped.

## Why

Blume generates two outputs that contradict each other.

`@astrojs/cloudflare` appends every `type: "redirect"` route to `dist/client/_redirects` with its exact status. **That file is the only thing that serves a redirect with its configured status, and only Cloudflare's static layer reads it.** Meanwhile `run_worker_first` is derived from the *first path segment* of the content routes, so one page at `/docs/reference` emits `/docs/*` — which swallows `/docs/api`, `/docs/api/run-query`, and every other redirect under `/docs`. Inside a worker-first route the static layer never runs, `_redirects` is never consulted, and Astro answers instead:

```js
// astro/dist/core/redirects/render.js
function computeRedirectStatus(method, redirect, redirectRoute) {
  return redirectRoute && typeof redirect === "object"
    ? redirect.status
    : method === "GET" ? 301 : 308;
}
```

`redirectRoute` is `routeMap.get(destination)`, keyed by *route patterns*. Blume serves every content page from `[...slug].astro`, so a destination never resolves to a discrete route, `redirectRoute` is always `undefined`, and a GET falls back to 301. Under Blume this is the guaranteed outcome, not an edge case — confirmed in the shipped manifest, where the redirect entry carries `"redirect": { "status": 302, ... }` and no `redirectRoute` key at all.

`buildRunWorkerFirstRules` only received `routePaths`, so it structurally could not know a redirect lived under a content prefix — but at the call site `config.redirects` was already in scope and simply unused.

## What this does

Passes the configured redirects into `buildRunWorkerFirstRules` (based the way the host platform matches them, via `applyBaseToPlatformRedirects`) and emits a negative rule for every one a positive rule would otherwise claim. That is the same mechanism already used to keep the raw `.md`/`.mdx` URLs on the static layer — for a stronger reason: `_headers` is cosmetic on a worker-first route, `_redirects` is load-bearing.

Things handled deliberately rather than incidentally:

- **At least one positive survives.** Wrangler rejects an all-negative list (`Only negative run_worker_first rules were provided; must provide at least 1 non-negative rule`). Exemptions are only ever derived from paths a positive rule claims, so a positive always exists; asserted in a test.
- **Redundancy.** Wrangler rejects a rule another same-polarity glob covers (`validateStaticRoutingRules`, applied per polarity). A collapsed section therefore pairs `!{path}` with `!{path}/*`, never `!{path}/`, which would be swallowed. Blume's own `coveredBy` mirrors this and leaves the new negatives intact — the `injectWorkerNegotiation` tests run through `mergeRunWorkerFirstRules`, so this is covered rather than assumed.
- **Rule ceiling.** A retired section collapses to a `{path}` + `{path}/*` pair — two rules however many URLs it held (11 → 2 on the site I verified against) — but only when no content route lives underneath, since the glob would otherwise take that route off the Worker and break its negotiation. Exemptions also ride along into `FALLBACK_RULES`, where `/*` claims every path and they matter *more* rather than less. If even that cannot fit the 100-rule cap, the negotiation is skipped entirely: a redirect cached as permanent is a worse outcome than raw Markdown staying reachable at its explicit `.md` URL.
- **Ordering.** Wrangler partitions the list into `assetWorkerRules` (negatives) and `userWorkerRules` (positives) before sending it, so relative order between polarities carries no meaning; the new negatives sit with the existing ones.
- **Content routes are never exempted.** The page owns the path, and taking it off the Worker would silently disable its Markdown negotiation.
- **Redirects no rule would have claimed get no rule**, so nothing is spent on paths already on the static layer.

## Test evidence

7 new tests, all failing before the change and passing after; full suite **2307 pass / 0 fail**, typecheck and `ultracite check` clean.

The discriminating test isolates the defect from any question about the redirect itself — same bundle, same `_redirects`, only the exemption differs. Verified end to end with `wrangler dev` on a real generated bundle (a site with `basePath: "/docs"`, one content page at `/docs/reference`, and 11 retired `/docs/api/*` redirects at `302`):

| `run_worker_first` | `/docs/api` | `/docs/api/run-query` | `/docs/api/v1/query/post` |
| --- | --- | --- | --- |
| `["/","/docs","/docs/*","!/docs/*.md","!/docs/*.mdx"]` — before | **301** | **301** | **301** |
| + `["!/docs/api","!/docs/api/*"]` — derived by this PR | **302** | **302** | **302** |

Markdown negotiation, which is the entire reason `run_worker_first` exists (#166), is unaffected on that same bundle: `Accept: text/markdown` still returns `text/markdown; charset=utf-8` on `/docs/quickstart` and `/`, `Vary: Accept` and `x-markdown-tokens` are still stamped, the homepage `Link` header survives, HTML requests still get `text/html`, and the root-level `_redirects` short links (`/twitter`, `/sitemap.xml`, `/openapi.json`) still serve `302`.

On that site the 11 redirects collapse to exactly the two rules its owner had been maintaining by hand in `wrangler.jsonc`, so the workaround in #169 becomes redundant — the generated set is identical, and because `mergeRunWorkerFirstRules` dedupes, leaving the manual rules in place produces byte-identical output rather than duplicates.

A changeset is included (patch).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Cloudflare routing for configured redirects, including nested paths, root redirects, and trailing-slash variants.
  * Preserved redirect status codes and content-route handling across deployments with configured base paths.
  * Prevented conflicting or unclaimed redirects from affecting unrelated routes.
  * Added safer fallback behavior when routing rules approach platform limits.

* **Documentation**
  * Added release documentation describing the updated Cloudflare redirect behavior.


haydenbleasel:
Thanks @akshatagarwl — great find and a thorough writeup; the `wrangler dev` verification table made this easy to trust.

I ran a deep review over the branch and pushed 16fb823 on top, which keeps your mechanism but hardens the edges the review surfaced:

- Exemptions are now derived from the **merged** rule set, so redirects claimed by a user's own `run_worker_first` rules are exempted too (a bare `true` is rewritten to its `/*` array spelling so the negatives can ride along).
- The content-route guards are based with `deployment.base` before comparing, so subpath deploys keep the "never exempt / never collapse over a content route" invariants — `redirectPaths` carry the full `{deployment.base}{basePath}` stack while the routes carry only `basePath`, so they previously compared disjoint namespaces off root deploys.
- The manifest routes (rather than `routePaths`, which always carries the synthesized llms.txt `/` mirror) guard the exemptions, so a configured root redirect can be exempted.
- An exemption over Wrangler's 100-character rule limit collapses to its nearest safe ancestor glob instead of skipping negotiation site-wide; with no safe ancestor your deliberate skip-negotiation ordering still applies.
- Only the request spellings a positive rule actually claims are emitted, conserving the 100-rule cap.

Plus small cleanups: a shared `normalizePath`, a `platformRedirects(config)` helper, and hoisted loop work. The exemption tests moved to `injectWorkerNegotiation` (where the exemptions now live) and grew to cover each of the cases above.

Merging — thanks again for the contribution!

PR #177, opened 2026-08-10
fix(openapi): keep diacritics in tag slugs

adrianbruegger:
## Summary
- OpenAPI tag slugify used `/[^a-z0-9]+/`, so letters like `ö`/`ñ` became hyphens
- Nav labels are humanized from that slug → `Größe` rendered as `Gr E`
- Slugify now keeps Unicode letters/numbers (`\p{L}` / `\p{N}`)

## Test plan
- [x] `bun run check`
- [x] `bun run typecheck`
- [x] `bun run test:coverage`
- [x] Docs build + `apps/docs` dev smoke
- [x] Unit: `Größe` / `Niños` keep diacritics; punctuation-only tags still fall back to `operations`

## Summary by CodeRabbit

* **Bug Fixes**
  * OpenAPI tag slugs now preserve Unicode letters and numbers, including accented and non-Latin characters.
  * Humanized labels no longer separate diacritics from their associated words.
  * Punctuation-only tags now use a consistent fallback slug, including collision handling.
  * Sidebar labels now preserve authored OpenAPI tag names and casing.
  * Canonical and social image URLs now correctly encode Unicode routes.
  * MCP server names now generate ASCII-compatible identifiers, with a safe fallback for unsupported names.

### DIFF
--- a/.changeset/six-sites-hammer.md
+++ b/.changeset/six-sites-hammer.md
@@ -1,5 +1,9 @@
 ---
-"blume": patch
+"blume": minor
 ---
 
-Keep Unicode letters in OpenAPI tag slugs so diacritics are not treated as word separators. Labels humanized from those slugs no longer split words like `Größe` into `Gr E`.
+Keep Unicode letters in OpenAPI tag slugs, and label tag sidebar groups with the spec's own tag names.
+
+Slugs derived from OpenAPI tag names (and reference-source labels) now keep Unicode letters and numbers instead of stripping them to hyphens, with NFC normalization so canonically equivalent spellings share one slug. Sidebar groups for tag sections now take their label directly from the spec's `tags[].name` (overridable with a `meta.ts` title), so authored casing like `OAuth2`, `iOS SDK`, or `Größe` renders verbatim instead of being re-humanized from the slug. Canonical and `og:image` URLs percent-encode route-derived paths, matching the sitemap. The MCP server card's reverse-DNS `name` now transliterates to ASCII (`Café Docs` → `cafe-docs`) to stay within the registry schema.
+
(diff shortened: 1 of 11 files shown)


adrianbruegger:
Addressed the CodeRabbit note: `NON_SLUG` now keeps Unicode combining marks (`\p{M}`) in both OpenAPI slug filters, with an NFD regression case for `Niños`.

haydenbleasel:
Thanks for the fix — the slug change itself was right, and it's merged into this branch unchanged. A review here turned up some follow-up issues in its blast radius, so I've pushed a commit (63d03cc4) on top of your branch (via maintainer edits) rather than round-tripping them through review comments:

- `slugify` now NFC-normalizes, so canonically equivalent spellings (e.g. NFD `Niños` from macOS tooling) land on one slug instead of two pixel-identical routes; drops format characters like ZWNJ instead of hyphenating them (`نرم‌افزار` stayed word-split under the new regex); and trims leading combining marks so a marks-only tag still falls back to `operations`.
- `model.ts` now re-exports the `references.ts` slugify instead of carrying an identical private copy — this PR had to make the same edit twice, which is how the two would eventually drift.
- Tag sidebar groups now take their label from the spec's own `tags[].name` (via source-contributed folder meta, overridable in `meta.ts`), so `OAuth2`/`iOS SDK`/`Größe` render verbatim instead of being re-humanized from the slug.
- Canonical and `og:image` URLs percent-encode route-derived paths, matching the sitemap's encoded `<loc>` — a raw `/api/größe` URI isn't legal and mismatched the sitemap byte-for-byte.
- The MCP server card's reverse-DNS name kept using the shared slugify, which would have published a schema-invalid Unicode identifier; it now transliterates to ASCII (`Café Docs` → `cafe-docs`, falling back to `docs`).
- The changeset is bumped to minor and documents that operation URLs move for non-ASCII tags, with a pointer at config `redirects` for already-deployed sites.

Known limitation left as-is: Turkish `İ` lowercases to `i` + combining dot, which has no precomposed form, so `İstanbul`/`Istanbul` still produce visually identical but distinct slugs. All checks pass (`check`, `typecheck`, full suite under the coverage gate).

PR #178, opened 2026-08-10
Split segmented index terms on punctuation held inside a word

h-kono-it:
## Description

A follow-up to #132. The segmenting tokenizer decides what to do with a segment by testing the whole segment, and that assumes a word-like segment contains only letters. UAX #29 does not work that way: connector punctuation, combining marks, format characters and mid-number punctuation are all held *inside* a word, so `Intl.Segmenter` reports each of these as one word-like segment.

```
スネーク_ケース   → one segment (WB13a/13b join Katakana across U+005F)
1,000            → one segment (WB12 joins Numeric across a comma)
robots.txt       → one segment
```

Each fails the all-script test, so it was indexed exactly as it stood. The punctuation then has to be retyped for the term to match, and the tail of the word is out of reach entirely — a query for `ケース` cuts to `ケー` / `ース`, and neither prefix-matches the segment the connector held together. A separate case: a few symbols are word-like on their own (U+00B8 among them), and those became index terms of their own.

The fix splits a word-like segment on anything that is neither a letter nor a digit, and treats that boundary as the end of a bigram run — the same thing a space or an interpunct already did, now applied when the separator sits *inside* the segment rather than between two of them. No window spans the punctuation, so `クーリング・オフ` and `スネーク_ケース` are cut alike.

On a 65-page Japanese corpus:

| Query | Before | After |
| --- | --- | --- |
| `ケース` (page has `スネーク_ケース`) | no hits | the page |
| `txt` (page has `robots.txt`) | no hits | 3 pages |
| `js` (page has `Node.js`) | no hits | 1 page |
| `¸` | 1 page | no hits |
| `スネーク_ケース`, `robots.txt`, `deployment.base` | hit | unchanged |

The 20-query benchmark from #132 is unchanged at 53/53 top-3 and MRR 1.000, so this only reaches text the old path could not.

Two notes on scope. The split applies to every segmented locale, not just the bigrammed ones — Orama's default tokenizer already splits Latin text on this punctuation, so `ja`/`zh`/`ko`/`th` were the odd ones out rather than the other way round. And halfwidth katakana (`ｻｰﾊﾞｰ`) and fullwidth Latin (`Ａｓｔｒｏ`) still fall through to the whole-segment path; that is a normalization question, not a punctuation one, and I left it alone here.

## Related Issues

Follow-up to #132. No open issue.

## Checklist

- [x] I've reviewed my code
- [x] I've written tests

## Additional Notes

Two tests added to `packages/blume/test/mcp.test.ts`, next to the existing bigram cases. `bun test --coverage` keeps `src/search/orama-index.ts` at 100%. `configured-integrations.test.ts` fails for me on a clean checkout of `main` as well, so I don't think it's from this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved multilingual search for punctuation, connector characters, and combining marks.
  * Preserved searchable terms containing apostrophes, version numbers, and decomposed characters.
  * Prevented punctuation from incorrectly combining unrelated words into search phrases.
  * Excluded symbol-only segments from search results.

* **Tests**
  * Added coverage for Japanese, Thai, mixed-script content, punctuation, combining marks, apostrophes, and numeric terms.


haydenbleasel:
Thanks for this — the restructure and the underscore/interpunct behavior were exactly right, and the tests pinned the intended cases well. While reviewing I found that splitting on `[^\p{L}\p{N}]+` also removed combining marks, which regressed scripts that spell with them: Thai vowel and tone signs were stripped (เสื้อ *shirt* and เสือ *tiger* collapsed to the same tokens, with the wrong page ranked first), and decomposed kana lost its voicing (パ indexed as bare ハ). Mid-number punctuation and apostrophes also over-split: `1.0.3` became the bag {1, 0, 3} and `don't` leaked a stray `t` token.

I pushed f52789e3 to this branch rather than round-tripping through review comments:

- Terms are now runs of letters, **combining marks**, and digits, with two survivors where the surrounding text makes them part of the word: an apostrophe followed by a letter, and `.`/`,` flanked by digits.
- Input is NFC-normalized before segmenting, so decomposed content matches composed queries (a mismatch that predated this PR).
- The "separator ends a bigram run" rule now lives in one `matchAll` loop instead of three spots.
- Added regression tests for all four behaviors; your original tests all still pass unchanged.

Everything the PR set out to fix still works — `スネーク_ケース` splits at the connector, `txt` reaches `robots.txt`, and the lone `¸` indexes nothing.

PR #187, opened 2026-08-14
Measure audit title and description limits in display columns

h-kono-it:
## Description

`BLUME_AUDIT_TITLE_LENGTH` and `BLUME_AUDIT_DESCRIPTION_LENGTH` compare `.length` against 10–60 and 110–160. What a search engine truncates is the space the text takes up, so a character count only stands in for that where every character is one column wide — true of Latin text and of nothing else.

Counted in characters, one range can't serve both scripts, and it misses in *both* directions at once. On the Japanese docs site I maintain (65 pages):

| Check | Now | This PR |
| --- | --- | --- |
| `DESCRIPTION_LENGTH` | 65 of 65 pages | 0 |
| `TITLE_LENGTH` | 0 | 10 |

The descriptions are ordinary Japanese ones of ~60–70 characters, which render about as wide as a 130-column English description — squarely inside the range, but reported as "too short" on every page. The titles that now report are 31+ fullwidth characters, wider than 60 columns, and they do truncate in a SERP. So the unit was simultaneously too strict for descriptions and too loose for titles; nudging the numbers can't fix both.

Measuring display columns instead, via `string-width`. A fullwidth or wide character counts 2, everything else 1, so **Latin text scores exactly as it did before** — an English site's findings don't move.

## Does the range still hold up?

I pulled the live `<meta name="description">` from real documentation sites and measured how many land inside 110–160 — 15 English (react.dev, MDN, Astro, Next.js, TypeScript, Tailwind, Stripe, GitHub, useblume.dev…) and 24 Japanese (MDN ja, Astro's own Japanese docs, Zenn, Qiita, and several government sites).

| | Characters (now) | Columns (this PR) |
| --- | --- | --- |
| English, 15 sites | 27% | 27% |
| Japanese, 24 sites | 13% | **29%** |

Ahrefs' range is opinionated — only about a quarter of well-maintained English sites sit inside it, and that's the check working as intended. The point is that a Japanese site scored 13% against the same bar and now scores like an English one. Small samples, so I'd read 27 vs 29 as "the gap closes", not as an exact match.

Two Japanese examples, for shape: Astro's own `docs.astro.build/ja` pages run 29–32 characters (58–64 columns), and MDN ja's `Array` page is 88 (170 columns).

## Related Issues

No open issue. This is the same class of thing as 051bceb (`Count audit prose words with Intl.Segmenter`, 1.4.3) — a Latin-calibrated unit reporting on every page of a CJK site — one file over from where that landed.

## Checklist

- [x] I've reviewed my code
- [x] I've written tests

## Additional Notes

`string-width` is a new dependency (MIT, ESM, sindresorhus). JavaScript's regex property escapes cover General_Category, Script and Script_Extensions but not `East_Asian_Width`, so the alternative is carrying a code-point table.

Three tests added to `audit-checks.test.ts`, including one asserting Latin text is scored identically so the no-change-for-English claim is pinned. `bun test --coverage` keeps `src/audit/checks/content.ts` and `src/audit/types.ts` at 100%; the full suite passes (2,450).

Two things I left alone deliberately: `minWordCount` is already segment-aware after 051bceb, and the thresholds stay non-configurable, since the comment in `types.ts` says that's waiting on the ids to settle. One behavior change beyond width: an emoji in a description used to count as its UTF-16 code units (a ZWJ sequence could score 8), and now counts as 2.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audit title and meta-description limits to measure displayed width, including fullwidth characters such as Japanese text.
  * Preserved existing behavior for Latin and ASCII content while handling combining marks and collapsed whitespace more accurately.
  * Updated generated release and OpenAPI descriptions to truncate safely by display width and retain ellipses and required suffixes.
  * Updated findings to report display-column widths more accurately.

* **Documentation**
  * Clarified length thresholds and how wide characters are measured.


haydenbleasel:
Thanks for this — the diagnosis is exactly right, the measurement data made it easy to accept, and pinning the ASCII path with a test was appreciated. I pushed 5eb38b3 onto your branch (the PR had maintainer edits enabled) to extend the fix rather than round-trip through review comments; here's what it does and why.

**The two generators that derive meta descriptions still budgeted in characters.** Both the changelog source (`github-releases.ts`, whose comment explicitly promises alignment with the audit thresholds) and the OpenAPI renderer (`render-mdx.ts`) cap their generated `seo.description` at 160 UTF-16 characters. With the audit now grading in columns, a Japanese release summary of ~120 characters sails under the character cap while rendering ~240 columns — so every generated changelog/API page on a CJK site would report `BLUME_AUDIT_DESCRIPTION_LENGTH`, and the finding's advice ("rewrite the frontmatter description") is impossible on a machine-generated page. That would have reintroduced the exact "finding on every page" failure this PR exists to fix, one layer down. Both generators now cut by columns via a shared grapheme-safe `columnsPrefix` helper (`src/core/text-width.ts`, generalized from your `graphemePrefix`); as a side effect the OpenAPI `clip()` is no longer able to split a surrogate pair. ASCII output is byte-for-byte unchanged, and the changelog docs' promise that summaries are "trimmed to the search-snippet length blume audit checks for" holds again.

**Width is now measured on whitespace-collapsed text.** Extraction only trims the ends, so a `<title>` authored across indented source lines kept interior newlines and indent spaces — and a page at a threshold boundary could flip between pass and finding based on source formatting a SERP collapses away.

**The "Latin scores exactly as before" claim is scoped to ASCII/NFC.** A decomposed (NFD) combining mark counted 1 under `.length` and counts 0 under `string-width`, so a ~110-character NFD description near the floor can flip on upgrade. Arguably the new number is the more correct one, so the code stays; the changeset/JSDoc wording is softened and a test pins the NFD behavior alongside your ASCII one.

Two smaller things: the "display columns, not characters" doc comment now sits on all four threshold fields rather than just `titleMin`, and the `AuditThresholds` JSDoc notes that East-Asian-*ambiguous* characters (★ ※ ①) count 1 per `string-width`'s default even though CJK fonts render them fullwidth — doing better there needs the page's language threaded into the audit, which felt out of scope here but is worth a follow-up issue if it bites your site.

Full suite is at 2,454 passing with the coverage gate still at 100% on every touched file. Thanks again — this was a model bug report as much as a fix.

haydenbleasel (approved):
LGTM — the column-based measurement is the right unit, the evidence in the PR body made it easy to verify, and with the generators aligned in 5eb38b3 the whole pipeline now agrees on it.

PR #197, opened 2026-08-19
Fix blank sidebar on archived version pages with header tabs

nateships:
## Problem

With `versions` configured alongside `navigation.tabs`, every archived page renders an empty left sidebar.

A version tree's navigation `root` is versionized (`/v1.0`, or `/docs/v1.0` under a basePath) while tab paths stay in current-docs space (`/`, `/docs`). `sidebarForRoute`'s root-tab check is a plain equality (`tab.path !== root`), so on an archived page the root tab is misread as a *section* tab; `sectionChildren` finds no group at the tab's path inside the snapshot tree and the sidebar collapses to `[]`. The versioning docs say the archived sidebar renders unscoped by tabs, but the unscoped fallback is never reached.

## Fix

Treat a tab whose path contains the whole tree root as the root tab: `tab.path !== root && !isUnderPath(root, tab.path)`. Inside an archived tree the current-docs root tab is a strict ancestor of the versionized root, so scoping falls through to `withoutTabSections` and the snapshot's full tree renders, exactly like the localized (`/en`) and based (`/docs`) root-tab cases the check already handles.

Section-tab scoping, the empty-tab guarantee (no leaking other tabs' sections onto group-less tabs like the changelog), and the longest-prefix nesting behavior are unchanged — covered by the existing suite.

## Testing

- New regression test: archived version tree with root + section tabs, plain and under a basePath (fails on main, passes with the fix)
- `bun test`: 2582 pass, 0 fail
- Reproduced and verified end-to-end on a real versioned site (4 archived snapshots, tabs + basePath `/docs`): archived sidebars were blank before, render the full snapshot tree after

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed archived and versioned pages with configured header tabs so the full navigation sidebar displays correctly.
  * Improved navigation handling when localized paths, version prefixes, or base paths differ from tab paths.
  * Prevented header tabs from being incorrectly marked as current on archived pages.
  * Preserved sidebar groups at the root of versioned navigation trees.

### DIFF
--- a/.changeset/version-tree-sidebar.md
+++ b/.changeset/version-tree-sidebar.md
@@ -2,4 +2,4 @@
 "blume": patch
 ---
 
-Fix the sidebar rendering empty on archived version pages when header tabs are configured. A version tree's navigation root is versionized while tab paths stay in current-docs space, so the root tab was misread as a section tab that owns no group in the snapshot; it is now recognized as the root tab and the archived sidebar renders its full tree.
+Fix the sidebar rendering empty on archived version pages when header tabs are configured. A version tree's navigation root is versionized while tab paths stay in current-docs space, so the root tab was misread as a section tab that owns no group in the snapshot; it is now recognized as the root tab — via a single shared containment check used by sidebar scoping, tab-section pruning, and hoisting alike — and the archived sidebar renders its full tree. Header tabs also no longer claim `aria-current` on archived pages: they link back to the current docs, so none of them is the current page there.
--- a/.changeset/version-tree-sidebar.md
+++ b/.changeset/version-tree-sidebar.md
@@ -1,0 +1,5 @@
+---
+"blume": patch
+---
+
+Fix the sidebar rendering empty on archived version pages when header tabs are configured. A version tree's navigation root is versionized while tab paths stay in current-docs space, so the root tab was misread as a section tab that owns no group in the snapshot; it is now recognized as the root tab and the archived sidebar renders its full tree.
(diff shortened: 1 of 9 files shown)


haydenbleasel:
Thanks @nateships — the fix is right, and it matches the documented contract in the versioning docs (inside an archived tree the sidebar renders unscoped by tabs). Reviewing it surfaced a few adjacent issues with the same root cause, so I pushed a follow-up commit (5ee216b) on top rather than bounce it back:

- **One shared definition of "the root tab".** The equality test this PR exempts in `sidebarForRoute` also lived in `withoutTabSections` and in the tab-section hoist set in `core/navigation.ts`, so under a versionized root the root tab could still land in the prune/hoist sets (reachable when a slug-flattened `(group)` folder sits at exactly the unversionized prefix, e.g. `/docs` under root `/docs/v1.0`). All three sites now use a single `isRootTab` containment predicate exported from `core/navigation.ts`, which also folds away the now-redundant `tab.path !== root` clause in the guard.
- **`aria-current` on archived pages.** The root tab still claimed every archived route through `activeTabForRoute`'s spans-everything fallback, so the header marked "Docs" as the current page on every `/v1.0/...` route even though it links back to the current docs. The header, mobile drawer, and tab bar now use a `currentTabForRoute` helper that marks no tab current inside an archived tree.
- **Stale docs.** The `sidebarForRoute` docstring (and `Navigation.root`'s) still asserted that root and tab paths always share one path space — the premise this fix invalidates — so they now describe the versionized case too.
- Added tests for the un-pruned archived sidebar and the current-tab behavior, and extended the changeset.

Lint, typecheck, and the full coverage suite (2585 tests) are green locally.

PR #215, opened 2026-08-26
fix: support curly heading ids

callumflack:
## Summary

- accept trailing `{#custom-id}` heading markers as an alternative to `[#custom-id]`
- keep explicit ids consistent across rendering, TOC generation, and source validation
- recognize raw HTML element ids as valid fragment-link targets
- document the syntax and add a patch changeset

## Verification

- `bun test packages/blume/test/heading-markers.test.ts packages/blume/test/core.test.ts packages/blume/test/markdown.test.ts packages/blume/test/validate-command.test.ts` (260 pass)
- `bun run --filter blume typecheck`
- `bun run check`
- downstream PDPP strict build and link validation

The full local coverage hook reached 2,985 passing tests but failed on four pre-existing 5-second audit subprocess timeouts plus the existing locale-dependent Orama punctuation test. No affected test failed.

Assisted-by: AI

## Summary by CodeRabbit

* **New Features**
  * Added support for trailing `{`#custom-id`}` markers on headings.
  * Added fragment-link support for IDs on raw HTML elements.
  * Improved broken-link diagnostics to reference missing anchor targets.

* **Bug Fixes**
  * Anchor IDs inside fenced code blocks are no longer treated as page anchors.

* **Tests**
  * Added coverage for custom heading IDs, HTML anchors, and fragment-link resolution.


haydenbleasel (approved):
Thanks for this — the `{#id}` form and HTML-id anchors are both worth having. I pushed a follow-up commit (a43a804) to your branch rather than round-tripping review comments; the substance:

- **`{#id}` only compiled in `.md`.** In `.mdx` a bare `{…}` is a JSX expression (`Could not parse expression with acorn`), but the scanner and `blume check` accepted it, so the page failed at `dev`/`build` time. The scan now reports an unescaped `{#id}` in an `.mdx` page — or in a `.md` partial spliced into one, mapped back to the partial's line — as `BLUME_MDX_CURLY_ANCHOR`. The docs point at the escape `\{#id\}`, which both pipelines already resolve to the same marker and which compiles in both formats. The translator prompt now preserves the marker too.
- **HTML-id scanning was too loose and too narrow at once.** It counted ids inside inline code, HTML comments, `<Prompt>` blocks, and capitalized component props (`<YouTube id="…">` in our own docs became a valid anchor), while missing formatter-wrapped tags, unquoted ids, and `id={"…"}`. Detection now lives inside the existing heading walk (one fence-aware pass instead of two), masks inline code and comments, matches wrapped tags, and only counts lowercase tags. The `[^<>]` bound also removes the quadratic backtracking the old `[^>]*` had on long lines.
- `PageRecord.anchors` is required and deduplicated; the validate reference and changeset describe HTML ids as anchor targets.

Approving — merging once CI is green.

PR #217, opened 2026-08-26
fix(theme): green and orange text misses WCAG AA in light mode

b0dea:
Green and orange text doesn't pass WCAG AA in light mode.

Badges draw `text-<hue>-700` on a `bg-<hue>-500/15` fill. For green and orange that isn't dark enough. Computed from the resolved Tailwind palette over the default white background:

| badge | fill | text | ratio | |
| --- | --- | --- | --- | --- |
| green — GET, 2xx, `<Badge variant="success">` | `#d9f7e5` | `#008236` | **4.32:1** | ✗ |
| orange — PUT, 4xx, `<Badge variant="warning">` | `#ffe8d9` | `#ca3500` | **4.44:1** | ✗ |
| teal — RECEIVE | `#d9f5f2` | `#00786f` | 4.69:1 | ✓ |
| red — DELETE, 5xx | `#fedfe1` | `#c10007` | 5.17:1 | ✓ |
| blue — POST, 3xx | `#dfecff` | `#1447e6` | 5.71:1 | ✓ |
| purple | `#f3e3ff` | `#8200db` | 5.82:1 | ✓ |
| violet — SEND | `#eee5ff` | `#7008e7` | 6.00:1 | ✓ |
| yellow — PATCH | `#fcefcc` | `#894b00` | 6.02:1 | ✓ |

The bar is 4.5:1 because these are all small text — 12px in the operation header, 10.4px in the reference sidebar, down to 10px on `<Badge size="xs">`. None is bold enough or big enough for the large-text exemption.

PATCH is the interesting one: it's already on `-800`. So the darker step got reached for once and green and orange were left behind.

Moving both to `-800` gives 6.20:1 and 6.25:1. The fill doesn't change, so the badges look the same weight. Dark mode is fine already — `-300` on the same fill over `oklch(0.085 0 0)` is 9.69:1 or better everywhere — so I left it alone.

While measuring I found a worse one, so it's in here too: the **`deprecated` label** on an operation is 10px text at `orange-600` straight on the page background, **3.59:1**. In `Operation.astro` it sits in the same flex row as the method badge. `orange-700` takes it to 5.23:1.

Files touched, all carrying the same pairs:

- `openapi/MethodBadge.astro` — GET, PUT, and GraphQL `QUERY` which reuses GET's colors
- `layout/NavTree.astro` — the sidebar copy of the same table (this is the 10.4px one)
- `openapi/Responses.astro` — 2xx and 4xx status chips
- `content/Badge.astro` — the `success` and `warning` variants (filled only; the stroke table below draws on the page background where `-700` is 4.94:1 and 5.23:1, so it stays)
- `openapi/{Operation,GraphqlOperation,AsyncApiOperation}.astro` — the `deprecated` label

GET is the most common method in a reference and it shows up once per sidebar row plus once per operation header, so it was the failing one people saw most.

Second commit does the same for **callout icons**, which are meaningful UI (the icon names the callout's type) so they carry the 3:1 bar of 1.4.11 rather than the decorative exemption. Three missed it: `check`/`success` at 2.94:1 and `warning` at 2.95:1 against their own tints, both now `-700` (4.52:1 and 4.67:1), and `note`'s icon was muted-foreground at 70% over the muted surface, 2.68:1, now full strength at 4.64:1.

**One thing I didn't fix here**, because it's a token rather than a component: `<Callout>` body text is `text-muted-foreground` over the variant tint, and two variants miss AA in light mode — `danger` at 4.384:1 and `info` at 4.497:1. `--blume-muted-foreground` is 5.06:1 on the page background, so it clears the bar there and has nothing left to spend on a tint. That's #221.

`bun run check`, `bun run typecheck` and `bun run test:coverage` all pass. Changeset included. Nothing pinned the old class strings.

Found this running Blume on a ~450-page API-heavy site. Happy to split it up per file if you'd rather.

## Summary by CodeRabbit

- **Accessibility**
  - Improved light-mode contrast for badges, labels, response statuses, cards, and callout icons to better meet WCAG requirements.
  - Improved tip callout contrast against customizable accent colors.
  - Dark-mode styling remains unchanged.

- **Consistency**
  - Standardized colors, icons, and status styling across navigation, OpenAPI, GraphQL, cards, badges, and callouts.

- **Bug Fixes**
  - Improved handling of callout types, including consistent treatment of “check” callouts as success callouts.

- **Tests**
  - Expanded accessibility checks across navigation and component content in both light and dark themes.


haydenbleasel (approved):
The measurements check out — I recomputed every ratio from the Tailwind palette and Blume's tokens and got your figures exactly. Thanks for the careful pass.

I pushed a follow-up commit (c6d4449) to the branch rather than round-tripping on review comments:

- `<Card>` had its own copy of Callout's pre-fix icon table, so typed cards still missed the 3:1 bar; both now read one table.
- Stroked `<Badge>` variants are `bg-transparent` and inherit their container, so `-700` green sits at 4.47:1 on the muted surface — `success`/`warning` take `-800` filled or stroked.
- The method/status/admonition tables and the `deprecated` label class now live in `components/colors.ts` with the contrast rule stated once, replacing the hand-measured numbers that would go stale on a palette bump. `check` folds into `success` before the lookups.
- `test/colors.test.ts` locks the rule in, and the docs' axe contrast gate now covers the components page and the sidebar.

Merging once CI is green.

PR #221, opened 2026-08-26
fix(theme): give muted text enough contrast to survive a tint

b0dea:
Split out of #217 because it's a token rather than a component, so the blast radius is different and you may want to weigh it separately.

`--blume-muted-foreground` is set so it clears WCAG AA on the page background and nowhere else. At `oklch(0.54 0 0)` that's 5.06:1 on white — fine in isolation. But muted text is 14px body copy and the theme routinely sets it over a 10% tint, which costs a few tenths. Two callouts land under the bar as a result.

Measured against the resolved tokens, light mode:

| surface | at `0.54` | at `0.53` |
| --- | --- | --- |
| `danger` callout | **4.384** ✗ | 4.573 |
| `info` callout | **4.497** ✗ | 4.692 |
| `tip` callout | 4.512 | 4.707 |
| `note` callout | 4.573 | 4.770 |
| `check`/`success` callout | 4.627 | 4.827 |
| `warning` callout | 4.682 | 4.884 |
| page background | 5.061 | 5.280 |

`info` at 4.497 is the tell — it's not failing by a design mistake, it's failing because there was no headroom to spend. Any tinted surface eats the margin.

`0.53` clears every surface the theme puts muted text on, worst case 4.57:1 inside `danger`. It's a tenth of a step in lightness; the intent is headroom, not a darker look. Dark mode is a separate token and is untouched.

I found this measuring the callout icons in #217 and left it out of that PR deliberately — a component class change and a global token change aren't the same review.

`bun run check`, `bun run typecheck` and `bun run test:coverage` pass, and the docs build is green. Changeset included.

## Summary by CodeRabbit

- **Accessibility**
  - Improved light-theme contrast for muted text, meeting WCAG AA guidance.
  - Reduced the accent tint on tip callouts and cards for clearer text contrast.
  - Dark-theme colors remain unchanged.

- **Documentation**
  - Added release notes documenting the updated muted foreground color and improved contrast ratio.


haydenbleasel (approved):
Thanks — the 0.54 → 0.53 nudge is right, and I reproduced every row of your table to three decimals from the Tailwind palette. The one that didn't reproduce was `tip`: it only matches with a teal accent, which is what useblume.dev sets. Under the shipped default accent (`oklch(0.145 0 0)`, near black) `tip` is a 10% black tint and muted text measured 4.07:1 before and 4.25:1 after — the worst callout surface, and still failing.

Since the accent is user-configurable with no lightness floor, no fixed text lightness can cover that surface, so I pushed a commit on top that caps the tint instead: `tip` on Callout and Card drops from `accent/10` to `accent/6` (4.62:1 worst case on pure black, 4.65 default, 4.92 teal). I also corrected the changeset and token comment, which claimed every callout surface cleared, and trimmed the comment. Approving with that in.

Issue #200, opened 2026-08-19
Option to configure/disable "Open in chat" page action?


billyjanitsch:
**Is your feature request related to a problem? Please describe.**

Currently, the"Open in chat" page action and the list of providers within are [static](https://github.com/haydenbleasel/blume/blob/718b58f14dcfe4155572af90bc88c33826374589/packages/blume/src/components/layout/PageActions.astro#L137-L171). There's no way to configure them or disable the action entirely.

**Describe the solution you'd like**

Personally, I'd like a boolean option to disable this feature entirely, similar to `feedback: false`.

**Describe alternatives you've considered**

Maybe other users would prefer the ability to override the list of providers, though I'm not sure how the config would interact with i18n and provider logos.

haydenbleasel:
Shipped in e31ff0a — this will go out in the next patch release.

`ai.openInChat` now controls the action:

- `false` hides it entirely
- an array of provider keys (`"v0"`, `"chatgpt"`, `"claude"`, `"t3"`, `"scira"`, `"cursor"`) shows just that subset, in the order you list them
- `true` (the default) keeps the current full list

On your i18n question: subsetting turned out to have no i18n or logo interaction — brand names render verbatim and only the surrounding "Open in {name}" phrase is localized, so the configurable list reuses the existing logos and labels as-is. Custom user-defined providers (own name/URL/logo) are out of scope for now.

Docs: https://www.useblume.dev/docs/configuration/ai#open-in-chat

PR #222, opened 2026-08-27
feat: Localize the "Open in chat" prompt

tykid16:
## Problem

The "Open in chat" action pre-fills the chat provider with `Read <url> so I can ask you questions about this page.` — hardcoded in English in `PageActions.astro`'s client script, regardless of the site's locale. The chrome around it is already dictionary-driven (`actions.openIn`, `actions.openInChat`), so a Japanese site shows 「チャットで開く」 and then hands Claude/ChatGPT an English prompt.

I hit this on the Japanese-only docs site I run on Blume (docs.irbank.net). Today we post-process the built `dist/` to swap the sentence for a Japanese one — which works, but breaks on any upstream wording change and isn't available to `blume dev`.

## Change

The prompt moves into the UI dictionary as `actions.openInChatPrompt`, the same route the other formerly-hardcoded chrome strings took:

- **Schema** (`core/i18n-ui.ts`): the new key defaults to the current English sentence, parameterized with `{url}` — the placeholder convention `actions.openIn` (`{name}`) and `changelog.showReleases` (`{version}`) already use. English behavior is byte-for-byte unchanged.
- **PageActions**: the string reaches the client script via a `data-i18n-chat-prompt` attribute (the `data-i18n-copied` pattern), with the English fallback inline. `{url}` is replaced at load time with the page's raw-Markdown URL — via a replacer function rather than a replacement string, since URLs may legally contain `$`, which string replacement would interpret as a substitution pattern.
- **Packs**: all 36 shipped starter packs carry a translation. Each follows the pack's established register (vous in `fr`, du in `de`, formal in `ru`/`uk`/`sr`, Cyrillic where the pack is Cyrillic). In languages where a first-person "so I can ask" would force a speaker gender (e.g. Polish _żebym mógł/mogła_), the sentence is rephrased to a gender-neutral construction. Happy to adjust any of these — they're starter-pack quality, written by a non-native speaker for most locales.
- **Test**: the UI-dictionary suite gains the existing per-pack pattern — every pack must carry the key, and every translation must keep the `{url}` placeholder.
- **Docs**: a note in `configuration/ai.mdx`'s Open in chat section pointing at the dictionary key and the `{url}` placeholder, linking to the Translated UI section.
- **Changeset**: `patch`.

## Related Issues

No open issue. Closest prior art is #200, which made the provider list configurable and localized the labels around the action — the prompt itself was the piece left hardcoded.

## Verification

- `bun run check`, `bun run typecheck`, and `bun run test:coverage` pass locally (macOS, Bun 1.3.14): 2987 tests, 0 fail, `packages/blume` stays at 100% per-file line and function coverage.
- `test/i18n.test.ts` (55 pass) carries the new per-pack assertions; removing the key from any pack fails it by name.

## Checklist

- [x] I've reviewed my code
- [x] I've written tests

## Summary by CodeRabbit

- **New Features**
  - Localized the “Open in chat” prompt across supported languages.
  - The prompt now automatically includes the page’s raw-Markdown URL.
  - Custom UI translations can override the default wording.

- **Documentation**
  - Documented prompt localization, customization, and `{url}` replacement behavior in the AI configuration guide.

- **Bug Fixes**
  - Removed the previously hardcoded English prompt when opening pages in chat.

### DIFF
--- a/.changeset/open-in-chat-prompt-i18n.md
+++ b/.changeset/open-in-chat-prompt-i18n.md
@@ -1,0 +1,5 @@
+---
+"blume": patch
+---
+
+Localize the "Open in chat" prompt. The prompt handed to the chat provider ("Read `<url>` so I can ask you questions about this page.") was hardcoded in English regardless of the site's locale. It now comes from the UI dictionary as `actions.openInChatPrompt` — with `{url}` replaced by the page's raw-Markdown URL at load time — so localized sites send a localized prompt, every shipped starter pack carries a translation, and `i18n.ui` can override the wording like any other chrome string.
--- a/apps/docs/content/docs/configuration/ai.mdx
+++ b/apps/docs/content/docs/configuration/ai.mdx
@@ -117,6 +117,8 @@ The **Open in chat** action opens the current page in an AI assistant — v0, Ch
 
 Like Copy as Markdown, it needs no setup. The assistant fetches the page over its public URL, so it works as soon as the page is deployed.
 
+The prompt is part of the [UI dictionary](/docs/content/i18n#translated-ui) (`actions.openInChatPrompt`), so localized sites send it in their language, and `i18n.ui` can override the wording — keep the `{url}` placeholder, which is replaced with the page's raw-Markdown URL.
+
 To tailor the action, set `ai.openInChat`. `false` hides it entirely, and an array of provider keys — `"v0"`, `"chatgpt"`, `"claude"`, `"t3"`, `"scira"`, `"cursor"` — shows just those providers, in the order you list them:
 
 ```ts blume.config.ts lineNumbers
--- a/packages/blume/src/components/layout/PageActions.astro
+++ b/packages/blume/src/components/layout/PageActions.astro
@@ -95,6 +95,7 @@ const menuRowClass =
 <div
   class="mt-8 space-y-0.5 border-border border-t pt-4"
   data-blume-page-actions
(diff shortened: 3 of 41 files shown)


haydenbleasel:
Thanks for this — clean change, and moving the prompt into the UI dictionary is the right shape for it. I ran lint, typecheck, and the coverage suite locally on the branch (GitHub Actions didn't kick off on this PR) and everything passed.

I pushed one commit to your branch (a203668) with two small follow-ups:

- **ko / zh / zh-tw:** the translations had text sitting directly against `{url}` — the Korean particle (`{url}을(를)`) and the full-width comma (`{url}，`). The old English template always had a space after the URL, and a chat provider's link detection can swallow those characters into the page URL. I reworded the three so the URL is followed by a space, and added a per-pack assertion in the test so it's caught next time. If you have a more natural phrasing for any of them, feel free to adjust — the test only cares that `{url}` is followed by end-of-string, whitespace, or ASCII punctuation.
- **PageActions.astro:** dropped the client-side `|| "Read {url} so I can…"` fallback. The frontmatter already merges `EN_UI.actions` under any override, so the attribute is always stamped and the English default now lives only in `core/i18n-ui.ts`.

One thing I'm deliberately leaving alone: nothing validates that an `i18n.ui` override keeps `{url}` (same as `{name}` and `{version}` today). The docs note to keep the placeholder, and that's fine for now.

PR #248, opened 2026-09-09
Stop the llms.txt audit flagging the MCP route as stale

b0dea:
`src/ai/llms.ts` lists the MCP endpoint in `llms.txt` whenever the server is on:

```
- [MCP server](${url(config.ai.mcp.route)}): Streamable HTTP Model Context Protocol server with ...
```

But that endpoint is a server route, not a file the build writes — `astro/generate.ts` puts it under `pages/api`, and the config comment there is explicit that "every Blume HTML route prerenders (the only server routes are API…)". So it lands in neither `context.byUrl` (page snapshots) nor `context.files` (the static output index), which is exactly the pair the stale-entry check tests:

```ts
if (!context.byUrl.has(path) && !context.files.has(path)) { …STALE_ENTRY }
```

The result is that **every server-output site with `ai.mcp` enabled** reports `BLUME_AUDIT_LLMS_TXT_STALE_ENTRY` against its own generated index, for a file that was never meant to exist. Under `--fail-on warning` that fails the audit outright and blocks publishing — which is how we hit it: our site has carried a local patch for this since 1.6.0 rather than being able to run the gate as shipped.

## The fix

The configured route is exempt while the server is enabled, and only then. With `ai.mcp` off, a listed `/mcp` is as stale as any other dead entry — there's a test for each direction.

Scope is deliberately one route. I checked every target the generator emits (`llms-full.txt`, `/index.md`, `agent-readability.json`, `sitemap.xml`, and `/.well-known/mcp.json`) and they are all real files, so the MCP endpoint is the only one with this shape.

## Verification

Test written first against `main` and confirmed failing for the right reason:

```
error: expect(received).not.toContain(expected)
Expected to not contain: "LLMS_TXT_STALE_ENTRY"
Received: [ "LLMS_TXT_STALE_ENTRY" ]
```

`bun run check`, `bun run typecheck`, `bun run test:coverage` and `bun run build` green on bun 1.4.0: 3334 pass, 0 fail, 100% lines and functions.

`test/audit-support.ts` gained an `mcp` option on the shared context builder, defaulting to disabled so no existing test changes behaviour.

## Summary by CodeRabbit

- **Bug Fixes**
  - Audits now recognize enabled MCP routes as valid served endpoints across links, redirects, and `llms.txt`, preventing incorrect broken or stale warnings.
  - MCP routes remain reported as stale when MCP is disabled.
  - Directory-served `index.html` entries are now accepted consistently.
  - Existing checks for other targets remain unchanged.

- **Documentation**
  - Added release documentation describing the updated MCP route auditing behavior.


haydenbleasel (approved):
Thanks, this is the right fix and the tests pin the boundary well (exempt only while the server is on).

One follow-up I pushed to your branch in 2753e38: the hunk extended an inline copy of the "is this path served" predicate, which left three near-copies (`links.ts`, `llms.ts`, and the redirect target set in `run.ts`) disagreeing in both directions. `llms.ts` rejected a directory served from its `index.html` that the link check accepted, and the link check reported a page link to the MCP route that `llms.ts` now exempted. I lifted the predicate into `audit/url.ts` as `isServed`, put the served-route exemption there, and pointed both checks and the redirect resolver at it. No existing test outcome changed; two new tests cover a page link and a redirect landing on the MCP route.

Check, typecheck, and the coverage suite are green on the branch. Approving.

PR #250, opened 2026-09-10
Render a spec description as the Markdown it is

b0dea:
## The problem

An OpenAPI `description` is Markdown by specification — *"CommonMark syntax MAY be used for rich text representation"* — but the reference components print it with `set:text`. On a spec generated from code docstrings that is most descriptions, so a property shows its own `**asterisks**` and backticks, a numbered list renders as one long line, and because HTML collapses newlines every paragraph and list runs together into a single wall of text.

What makes it stand out is that the **operation** description on the same page is fine: it is emitted into the MDX body and goes through the full pipeline. Two paths, one page.

Before and after on the same property, from a real spec:

```
Inline LLM configuration embedded on the agent. Carries provider, credentials, model
and tuning controls as a single blob. ... Two credential shapes (XOR, mirrors
:class:`ValidateLlmRequest`): 1. **Inline** - ``apiKey`` is filled directly. The BE
KMS-encrypts it at save and decrypts it at deploy / test-run. 2. **Linked** - ...
```

becomes three paragraphs, a real `<ol>`, bold labels and code chips.

## What this changes

A new `Description.astro` renders the string, used by `SchemaProperty`, `ParametersTable`, `RequestBody` and `Authorization`. Deliberately **not** the playground and message-composer hints — those are inline `<span>` labels inside a form, where block markup does not belong.

`marked` rather than the full Markdown pipeline: it is synchronous, already a dependency, and already how the Ask AI island renders model Markdown, while the pipeline is async and built for whole documents. These are thousands of short strings per build — a large reference renders tens of thousands.

## Two things it deliberately does not do

I got both of these wrong first, and each was caught by `blume audit`, so they are worth calling out.

**It does not link a bare URL or a relative href.** A description is frequently prose that was never Markdown. GFM autolinks bare URLs, and spec descriptions are full of example hosts (`https://myorg.my.salesforce.com`, `https://yourstore.myshopify.com`) that became anchors to hosts nobody should visit. Worse, regex and format notation reads as link syntax: Debezium's column-list wording

```
schemaName[.]tableName[.](columnName1|columnName2)
```

is literally `[text](href)`. Rendering it produced a broken link on 48 pages **and deleted the notation**, because the link text is only `.`. So an anchor is emitted only for an absolute URL, root path, fragment or `mailto:`, and the demoted case renders `raw` rather than `text` — which is what keeps the notation intact.

Raw HTML is escaped rather than passed through, for the same reason the Ask AI island runs DOMPurify over model output: this is data lifted out of a file and interpolated into the page. DOMPurify needs a DOM, so it is unavailable in a component that renders on the server; escaping is the equivalent that works there.

**It does not style with utility classes.** That repeated the same ~620 characters on every description: on a page with 2,400 of them, 1.5 MB of identical class attributes, which pushed eight pages past Googlebot's 2 MB crawl limit — an indexability error from `blume audit`. One class and one style block cost the same at 2,400 descriptions as at one; that page went 3.36 MB → 1.88 MB. Plain CSS rather than `prose`, too: the typography plugin sets its own size, colour and rhythm, which fight the muted small type these sit in.

## Verification

I could not run `bun run check` / `bun test` locally — my bun is 1.3.5 against the repo's pinned 1.3.14 — so please let CI judge formatting, and I am happy to follow up on anything `ultracite` wants.

What I can offer instead is that this exact code is running against a real 448-page site with 266 generated reference pages: build clean, `blume validate --strict` no broken links, and `blume audit --fail-on warning` at **34,124 audits · 0 errors · 0 warnings**. Both regressions above were found by that audit rather than by review, which is a nice advertisement for it.

A changeset is included.

## Summary by CodeRabbit

- **New Features**
  - OpenAPI, AsyncAPI, and GraphQL descriptions now support Markdown formatting, including paragraphs, lists, links, inline code, and code blocks.
  - Markdown rendering applies consistently across schemas, parameters, request bodies, responses, authorization schemes, messages, fields, enum values, and operation return types.
  - GFM tables display in horizontally scrollable containers.
  - Intended links remain clickable, while unsupported link-like text, images, and raw HTML are safely displayed as text.

- **Performance**
  - Consolidated description styling reduces unnecessary page markup and helps keep documentation pages within crawler size limits.


haydenbleasel (approved):
Thanks — this is a good change, and the reasoning in the description (the Debezium notation case, the utility-class bloat caught by `blume audit`) is exactly the kind of context that makes it easy to merge.

I pushed one commit on top (f25a414) rather than sending it back:

- **Bare `www.` hosts and emails were still autolinked.** The `raw !== href` test only catches the `https://` autolink; for `www.example.com` and `ops@example.com` marked prefixes `http://` / `mailto:`, so the strings differ just as they do for a deliberate link. A deliberate link is now one written as a link (`[`-prefixed or an angle autolink) with an accepted href.
- **Images** are held to the same href policy, so `x` and relative sources demote to their source text like a link does.
- **Response descriptions and AsyncAPI message descriptions** were the two remaining plain-text sites; both go through `Description` now.
- Headings inside a description get label weight so they don't collapse into body text.

Verified locally with `bun run check`, `bun run typecheck`, and the full coverage suite (3376 pass), plus a sandbox build with a Markdown-heavy description injected into the spec. CodeRabbit's three findings were all already addressed in your later commits.

PR #251, opened 2026-09-10
Give a reference row room for what it holds

b0dea:
> Stacked on #250 — the diff shown here includes that branch. Reviewing this after #250 lands (or merging that first) gives the clean two-file diff.

## The row

A reference row is scaled for a one-line description:

| | before | after |
|---|---|---|
| property name → description | 4px | 8px |
| description → "Show properties" | 8px | 12px |
| between two properties | 25px | 33px |

Once #250 renders a description as the Markdown it is, it is a **block** — and at 4px it sat closer to the label above it than its own paragraphs sat to each other (8px). That inversion is what makes a dense reference page read as a wall rather than as rows. The disclosure below it touched the next row's divider.

The new scale is one step per unit: 8px separates a row's own lines, 12px separates the disclosure, 16px of row padding separates the rows. Applied to every component that draws a reference row — `SchemaProperty`, `ParametersTable`, `Authorization`, `GraphqlFieldsTable`, `GraphqlType`, `GraphqlOperation`, `Bindings` — because they are copies of the same row and half of them fixed reads worse than none.

## The table cell

`.blume-table-scroll :is(th, td)` was `0.5rem 0.75rem`. Against a table line-height near 1.7, a cell whose content wraps put **more space between its own two lines than between itself and the next row** — the same inversion, in the component a reference corpus uses most.

Block padding goes to `0.75rem`. The inline padding stays put deliberately: widening it also reads better in isolation, but it comes out of column width in a capped article, and measured on a 448-page corpus it pushed cells that fit on two lines onto three, spending the space it had just bought.

## How the numbers were taken

`getBoundingClientRect` in a headless Chromium against the built pages of a real 448-page corpus, before and after, at 1440px. Not read off the class names — the first pass at this looked right in the source and measured wrong, because the description's internal paragraph margin was the thing setting the floor.

Also verified downstream on that corpus, carrying this as a patch: `blume audit` 34,124 audits with 0 errors and 0 warnings — worth stating because the audit's 2 MB crawl-limit check is sensitive to markup on exactly these pages — 86 tests green, and `blume check --strict` clean.

https://claude.ai/code/session_01TKgxJk5u2kcfuzHtG2AVNh

## Summary by CodeRabbit

* **UI Improvements**
  * Improved spacing and readability across reference rows, OpenAPI details, GraphQL documentation, bindings, parameters, and schema properties.
  * Standardized descriptions and supporting content with a consistent vertical rhythm.
  * Increased table cell vertical padding while preserving existing horizontal spacing.
* **Bug Fixes**
  * Improved rendering consistency for formatted descriptions throughout API reference pages.

### DIFF
--- a/.changeset/reference-row-vertical-rhythm.md
+++ b/.changeset/reference-row-vertical-rhythm.md
@@ -1,0 +1,7 @@
+---
+"blume": patch
+---
+
+Give a reference row room for what it holds. The row was scaled for a one-line description — 4px between the property name and its description, 8px before the disclosure, 12px of row padding — but a description is a block, so at 4px it sat closer to the label above it than its own paragraphs sat to each other. That inversion is what made a dense reference page read as a wall rather than as rows, and the disclosure below it touched the next row's divider. Every component that draws a reference row now shares one scale: 8px between a row's own lines, 12px before a disclosure, 16px of row padding. Measured on one operation page, the gap between two properties goes from 25px to 33px.
+
+Table cells gain the same treatment for the same reason. At 0.5rem of block padding against a line-height near 1.7, a cell whose content wrapped put more space between its own two lines than between itself and the next row. The inline padding is unchanged on purpose: widening it comes out of column width in a capped article, and on one corpus it pushed cells that fit on two lines onto three, spending the space it had just bought.
--- a/packages/blume/src/components/openapi/Authorization.astro
+++ b/packages/blume/src/components/openapi/Authorization.astro
@@ -39,7 +39,7 @@ const { security } = Astro.props;
             {alternative.map((resolved) => {
               const carrier = schemeCarrier(resolved);
               return (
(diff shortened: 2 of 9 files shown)


haydenbleasel (approved):
Checked this out and ran the sandbox on main and on this branch side by side. The measured changes match the description: reference row padding 12px → 16px, a parameter row with a description 88px → 104px, a Markdown table row 39px → 47px. Rows read as rows now, and the scale is consistent across the seven components.

One note for the record: the table-cell padding change applies to every Markdown table on a site, not only reference pages. Approving with that understood.

haydenbleasel (approved):
Rebased onto main after #250 landed so the PR carries only its own commit. The spacing change reads well against the Markdown descriptions, and lint, typecheck and the full suite pass locally on the rebased branch. One note for the record: the table-cell padding change in the theme applies to every table on the site, not only reference tables.

PR #259, opened 2026-09-13
Lazy-load CLI commands

b0dea:
## What changed

- resolve Citty subcommands lazily so a command does not initialize unrelated command dependencies
- preserve those boundaries in the published Node CLI with Bun code splitting
- cover selected-command execution and help with a subprocess import guard

## Why

On a one-CPU Bun 1.4.0 run, `blume eval`'s invalid-flag subprocess test consistently exceeded Bun's unchanged five-second deadline because each subprocess initialized every CLI command. With this change, the same three-run reproduction completes in 797–860 ms per run.

## Validation

- mandatory pre-commit hook: formatting, typecheck, 3,388 tests across 161 files, coverage, and package build passed
- one-CPU `eval` reproduction: 3 passed, 0 failed; 797–860 ms per run
- one-CPU `validate` reproduction: 3 passed, 0 failed; 270–350 ms per run
- Node 22 packaged entry with Astro unavailable: selected `eval` validation and `eval --help` both behaved as intended
- package build: 90 outputs plus type declarations; the package's existing `files: ["dist", ...]` includes the generated chunks

## Summary by CodeRabbit

* **Performance**
  * Improved CLI startup time by loading only the selected command when it is invoked.
  * Reduced unnecessary loading of unrelated commands while preserving command execution and help output.

* **Bug Fixes**
  * Ensured CLI logging respects the actual environment configuration when project environment files define conflicting values.
  * Standardized `--host` handling for development and preview commands, including binding to all interfaces when specified without a value.

* **Documentation**
  * Updated CLI bundle documentation to reflect code-split command files loaded on demand.


haydenbleasel (approved):
Thanks — the mechanism is right and the win is real (on a Node build, `validate --help` went from ~870 ms to ~280 ms here). I pushed a follow-up commit (d00f521) onto this branch rather than round-tripping comments; the details, so you can see what changed and why:

**Logger vs. `.env` ordering (regression).** On main, `consola` was imported before `loadEnvFiles` ran, so a project `.env` couldn't influence it. With the commands loaded lazily, consola first evaluated *after* `.env` had been applied, so a project `.env` containing `CONSOLA_LEVEL=0`, `NODE_ENV=test`, `TEST=1`, or `CI=true` silenced or restyled every command's output under Node (`blume version` printed nothing). The entry now imports `./log.ts` ahead of `loadEnvFiles`, and a test builds the bundle and runs it under Node with such a `.env` to keep it that way.

**`preview` still loaded the dev graph.** `commands/preview.ts` imported the one-line `normalizeHost` from `commands/dev.ts`, which pulled Astro's dev server, chokidar, and the dev lock into every `blume preview`. It lives in `host-args.ts` beside `normalizeHostArgs` now.

**Root usage and typos loaded all 15 commands.** citty resolves every subcommand to render `blume --help` and to check aliases for an unknown name, so bare `blume`, `--help`, and `blume typo` were no faster than before (~1 s). Each command's `meta` now lives in a shared `command-meta.ts` table (the command modules import theirs from it, so nothing can drift), and `lazy-command.ts` defers only `args`/`setup`/`run`/`cleanup`. Root help and unknown-command paths now take ~60 ms.

**Chunk layout.** The split chunks landed at `dist/` root beside `dist/types/`. They're named under `dist/cli/` now, the `Bun.build` options moved to `scripts/cli-bundle.ts`, and a test builds with those options and asserts the entry's static import graph never reaches astro / vite / the MCP SDK — so dropping `splitting` can't ship eager loading silently again. Comments in `build.ts`, `bin/blume.mjs`, `package-root.ts`, and AGENTS.md that described a single-file bundle are updated.

**Tests.** The preload guard now rejects every command module except the selected one (it previously only rejected `dev.ts`, so an eager `build.ts` import in the entry passed), and covers `preview --help`, root usage, a bare `blume`, and an unknown command. The four-subprocess `blume eval` flag test has an explicit 30 s timeout.

One thing left as-is: if the selected command's own chunk fails to import (a broken optional dep), citty's error path re-resolves its args and the failure surfaces through the `BLUME_INTERNAL` handler. That only affects the command that is actually broken, which is an improvement over main crashing every invocation, so I didn't change it here.

PR #275, opened 2026-09-19
Add ai.ask.cors for cross-origin callers of the Ask AI endpoint

radjavi:
## Summary
- add `ai.ask.cors`, a list of origins allowed to call the generated `/api/ask` from another site (a marketing page that embeds an ask box, for example)
- the generated route answers the browser's `OPTIONS` preflight and names a listed origin on every response — the streamed answer and the 400/500 statuses alike — so a cross-origin caller can tell a rejected body from a provider failure; unlisted origins get no header
- entries are reduced to their origin in the schema (`https://www.example.com/docs/` → `https://www.example.com`), and the route is byte-identical to today when `cors` is unset
- wire it through `generate` and `eject`, cover schema/template/generate/eject in tests, and document it on the Ask AI page next to the external-endpoint contract it mirrors

## Test plan
- [x] `bun run check`
- [x] `bun run typecheck`
- [x] `bun run build`
- [x] sandbox app built with `cors` set and served with the node adapter: `OPTIONS` from a listed origin → 204 with the allow headers, from an unlisted origin → 204 without them; `POST` 400/500 responses carry `access-control-allow-origin`

## Summary by CodeRabbit

- **New Features**
  - Added `ai.ask.cors` support for approved websites, including the wildcard `"*"` option.
  - Added preflight handling and CORS headers for successful, streamed, and error responses.
  - Origins are validated and normalized; unlisted origins remain restricted.
  - Cross-origin requests require a JSON `POST` body with `content-type: application/json`.

- **Documentation**
  - Documented cross-origin configuration and request requirements.
  - Clarified that externally configured endpoints manage their own CORS settings and cannot be combined with `cors`.


haydenbleasel (approved):
Thanks, this is a clean addition and the wiring through generate, eject, the schema drift guard and the docs is all there. I pushed a follow-up commit (c91765e) to the branch rather than round-tripping, so this is approved as it stands now. What changed and why:

- **CORS moved into a runtime module, `blume/ai/cors.ts`.** The route now imports `preflightResponse` and `withCors` instead of inlining a header set and stamping it on each `return`. Wrapping the `POST` once means the next return site added to the handler (a 429, a 413) can't ship an opaque failure for that one status. The module is unit-tested directly instead of via string assertions on the emitted source.
- **`Vary: Origin` on both branches.** The answer depends on the `Origin` header whether or not it was listed, so a shared cache that stored the header-less 204 (the route advertises a 24h max-age) could serve it to a listed origin, or the reverse.
- **Preflight reflects `access-control-request-headers`**, falling back to `content-type`, so a caller's fetch wrapper can add a header without an eject.
- **`cors: ["*"]`** is accepted; the schema only took URLs and rejected the wildcard with a bare "Invalid URL".
- **`cors` alongside `endpoint` is now a config error.** No route is generated in that case, so the entry silently did nothing.
- **Docs call out the JSON content type.** Astro's cross-site request check 403s a cross-origin `POST` with no content type (or a form-like one such as `text/plain`) before the route runs, and that 403 carries no CORS headers, so the browser reports a network error rather than the status. Your sandbox test passed because it sent JSON; a caller writing `fetch(url, { method: "POST", body })` would hit it.

Verified on the sandbox app with the node adapter: `OPTIONS` from a listed origin gets the allow headers plus the reflected request headers, an unlisted origin gets `vary: origin` only, `POST` 400/500 carry `access-control-allow-origin`, and a `POST` without a content type gets Astro's 403.

One thing I left alone: the MCP route keeps its own wildcard header set, since it also needs the session-id headers and applies them to every response.

PR #276, opened 2026-09-19
Add ai.ask.reasoning to control how much the Ask AI model reasons

radjavi:
## Summary
- add `ai.ask.reasoning` (`none` | `minimal` | `low` | `medium` | `high` | `xhigh`), forwarded as the AI SDK's top-level `reasoning` option in the generated route's `streamText` calls (grounded and plain paths alike), so each backend maps it to its own control — OpenAI's `reasoning_effort`, for example
- for grounded docs Q&A the retrieved excerpts carry the answer, so a reasoning model's default thinking is mostly time-to-first-token the reader waits through; `reasoning: "none"` turns it off, and leaving the field unset keeps the model's default (the route is unchanged)
- wire it through `generate` and `eject`, cover schema/template/generate/eject in tests, and document it on the Ask AI page next to retrieval size, which trades along the same axis

## Test plan
- [x] `bun run check`
- [x] `bun run typecheck`
- [x] `bun run build`
- [x] sandbox app built with `reasoning: "none"` and served with the node adapter: the generated route bundles and answers `POST /api/ask`

## Summary by CodeRabbit

- **New Features**
  - Added configurable Ask AI reasoning levels: `none`, `minimal`, `low`, `medium`, `high`, and `xhigh`.
  - Applied configured reasoning to generated and ejected Ask AI endpoints using each provider’s supported controls.
  - When unset, Ask AI retains the provider’s default reasoning behavior.
  - Configurations using reasoning with Inkeep are rejected because it does not support reasoning controls.

- **Documentation**
  - Documented supported levels and provider compatibility requirements.

- **Tests**
  - Added coverage for valid, invalid, configured, and default reasoning settings.


haydenbleasel (approved):
Thanks for this, it's a well-scoped addition and the grounded-Q&A motivation is spot on. We'll merge it.

One heads-up on where it's headed: on the next major release we'll likely rework the Ask AI provider config into adapters (something like `provider: openrouter({ ... })`), so each provider owns its own option mapping instead of Blume translating fields like `reasoning` per backend. That will most likely be a clean break rather than a compatibility layer: the flat `ai.ask` provider props, `reasoning` included, would be replaced by each adapter's own options instead of kept alongside them.

I pushed a follow-up commit to the branch that fixes the findings from review rather than round-tripping them:

- **OpenRouter dropped the option silently.** `@openrouter/ai-sdk-provider` 3.0.0 never reads the AI SDK's top-level `reasoning` call option; it only sends its own model-level setting. The route now emits `openrouter(model, { reasoning: { effort } })` for that backend and keeps the top-level `streamText` option for the gateway and the OpenAI-compatible providers, which do forward it.
- **Inkeep** has no reasoning control, so the schema now rejects `reasoning` with `provider: "inkeep"` instead of sending `reasoning_effort` on every request.
- **Docs, JSDoc, and changeset** now say what each backend does with the level. Neither installed provider warns about an unsupported value, so the "ignores it with a warning" sentence is gone.
- Built the `streamText` argument list once for both paths, restored the `askAiProviders` doc comment the new constant displaced, and moved the reasoning assertion out of the shared kitchen-sink fixture into a focused generate test.

Verified locally with `bun run check`, `bun run typecheck`, `bun run build`, and `bun run test:coverage`.

PR #289, opened 2026-09-20
Report Ask AI usage through the configured analytics providers

radjavi:
## Summary
- `useAskAI` (and so the built-in Ask AI island and any custom chat UI on `blume/hooks`) now reports usage through the same `track()` helper page feedback uses, so questions land in the configured analytics providers next to pageviews: `ask` when a question is sent, `ask_answer` when the answer finishes streaming (`ms`, `chars`), `ask_error` when the request fails (`ms`, `status`, with `0` for no response). Each event carries the `question` and the `path` it was asked from.
- Clearing the conversation mid-answer reports neither outcome, matching the existing generation guard. A provider that throws is swallowed, so analytics can never turn an answer into the error notice.
- Documented on the Ask AI page under a new "Analytics" section; changeset included.

Motivation: with Ask AI on a docs site (and a marketing site calling it through `ai.ask.cors`), the docs panel's questions were invisible except as model-provider request counts.

## Test plan
- [x] `bun run check`
- [x] `bun run typecheck`
- [x] `bun run build`
- [x] hooks tests assert the three events and their properties on the answer, non-OK, thrown-fetch, empty-question, and mid-answer-reset paths, plus the throwing-provider guard; the island tests' window stub gained `dispatchEvent` for the helper's universal hook

## Summary by CodeRabbit

* **New Features**
  * Ask AI usage is reported through configured analytics providers.
  * Events include question length, completed answer length, response time, page path, and request status.
  * Interrupted or empty responses are reported as errors.
  * Question text remains available through universal tracking events but is not sent to analytics providers.
  * Analytics provider failures no longer disrupt other tracking or Ask AI responses.

* **Documentation**
  * Updated Ask AI analytics documentation with event properties and status behavior.


haydenbleasel (approved):
Thanks for this — the events, the reset guard, and the docs section are all in the right shape. I pushed one commit (e42f2cfd) on top with the fixes from review rather than round-tripping them:

- **Stream aborts report their real status.** `streamText` defers provider and credential errors to the stream, so the backend's most common failure is a 200 whose body breaks mid-flight. That was landing in the catch as `status: 0` ("no response"). The hook now keeps the status from the response, and a 200 that delivers nothing (no body, empty stream) reports `ask_error` instead of a zero-char `ask_answer`.
- **`path` matches feedback and pageviews.** The events keyed on the base-stripped route while `PageFeedback` and the providers record the served pathname, so nothing joined under a `base`. Events now carry `location.pathname`; the API body keeps the stripped route for grounding.
- **Question text stays off third-party providers.** It's free-form reader input (pasted keys, error logs) that breaches GA4/Vercel's PII terms and their per-value size caps (Vercel drops the whole event over 255 chars). Providers get `questionChars`; the text travels only on the `blume:track` DOM event via a new optional third argument on `track()` for local-only props, so a site decides where it goes.
- **Per-provider isolation moved into `track()`.** A throwing `gtag` shim was still starving Plausible and the custom event, and the feedback click handler had no guard at all. Each provider call is isolated now, so the hook's own try/catch and its test went away.
- **Latency uses rounded `performance.now()`**, matching the playground client, so a clock jump can't produce negative or huge `ms`.

Docs table, prose, and the changeset updated to match. Full suite green with both touched files at 100% coverage.

One thing I looked at and left alone: the static `track` import pulls `@vercel/analytics` into `blume/hooks`. The feedback widget in the root layout already loads that module on every page, so it's a shared chunk and the hook import adds nothing to a default build.

